PR #5, opened 2020-05-16
Add HOTH description

### LINKED ISSUE #4: Create HothDescription Component
Create HothDescription component as shown in this [figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=0%3A1).

jamieliu386:
Fixes #4


src/components/HomePage/HothDescription.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { Container, Typography, Grid } from '@material-ui/core';
+import { useTheme } from '@material-ui/core/styles';
+import useMediaQuery from '@material-ui/core/useMediaQuery';
+
+const description = `Hack on the Hill (HOTH) 7 is a 12-hour, beginner-friendly
```
TimothyGu, line 6:
```suggestion
const description = `Hack on the Hill (HOTH) is a 12-hour, beginner-friendly
```

src/components/HomePage/HothDescription.js
```diff
@@ -0,0 +1,38 @@
...
+            <Grid item sm={9} md={7}>
+              <Typography variant="p">{description}</Typography>
+            </Grid>
+          </Grid>
+        </Container>
+      </div>
+    );
+}
+
+export default HothDescription;
```
TimothyGu, line 38: Your text editor doesn't add a new line at the end of the file. You should fix your editor so that it does so correctly, but for now you can add a new line manually.

src/components/HomePage/HothDescription.js
```diff
@@ -0,0 +1,38 @@
...
+      <div>
+        <Container maxWidth="md">
+          <Grid
+            container
+            justify="center"
+            alignItems="center"
+            spacing={isSmall ? 1 : 6}
+          >
+            <Grid item sm={9} md={5}>
+              <Typography variant={isSmall ? "h4" : "h3"} align={isSmall ? "center" : "right"}>WHAT IS HACK ON THE HILL?</Typography>
```
TimothyGu, line 27: For accessibility, the text should be normally cased. The presentation aspect (turning letters into upper case) should be done using CSS – i.e., `className` and `useStyle`.

Also, I would add another prop `component="h3"`. This way the heading level of the HTML element doesn't get changed when one moves from a device to another, but the styling will still change (`variant`). By default, `variant` controls both the style and underlying component/element used, but if we explicitly specify a `component` then it will listen to us rather than `variant`.

TimothyGu:
Merge at will!

Issue #1, opened 2020-05-09
Create NavBar Component


Timthetic:
Create a NavBar Component as shown in this [figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=0%3A1).

PR #15, opened 2020-05-23
Responsive navigation bar

milesswu:
Fixes: #1


src/components/MenuBar/ButtonBar.js
```diff
@@ -0,0 +1,70 @@
...
+
+const styles = theme => ({
+    btn: {
+        fontFamily: theme.typography.fontFamily,
+        fontWeight: 500,
+    },
+});
+
+function ButtonBar(props) {
+    const { classes, isMobile } = props;
```
TimothyGu, line 17:
```suggestion
function ButtonBar({ classes, isMobile }) {
```

This gets rid of the `props` variable that we shadowed later in the Link components.

src/components/MenuBar/ButtonBar.js
```diff
@@ -0,0 +1,70 @@
...
+				to: "/"
+			},
+			{
+				name: "gallery",
+				to: "/"
+			}
+		];
+
+    return (
+        <React.Fragment>
```
TimothyGu, line 48:
```suggestion
        <>
```

src/components/MenuBar/ButtonBar.js
```diff
@@ -0,0 +1,70 @@
...
+            {links.map((link) =>
+								<PoppinLink to={link.to}>
+									{link.name}
+								</PoppinLink>
+						)}
+            <BorderLink to="/">
+                Apply
+            </BorderLink>
+        </React.Fragment>
```
TimothyGu, line 57:
```suggestion
        </>
```

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,118 @@
...
+			padding: '3%'
+		}
+	};
+};
+function MenuBar({ classes }) {
+	const [menuOpen, setMenuOpen] = useState(false);
+	const toggleMenu = () => setMenuOpen(open => !open);
+
+	return (
+		<React.Fragment>
```
TimothyGu, line 82:
```suggestion
		<>
```

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,118 @@
...
+			<AppBar position="sticky">
+				<Toolbar className={classes.toolbar}>
+					<div
+						className={classes.logohome}
+						onClick={() => navigate('/')}
+					>
+					</div>
```
TimothyGu, line 89: Could you add something here, like "ACM Hack", so we remember this is where to put the wordmark.

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,118 @@
...
+			{/* Mobile menu Bar */}
+			<div className={classes.mobileMenuBar}>
+				<Collapse in={menuOpen}>
+					<div className={classes.mobileBtnContainer}>
+						<ButtonBar isMobile/>
+					</div>
+				</Collapse>
+			</div>
+
+	</React.Fragment>
```
TimothyGu, line 110:
```suggestion
	</>
```

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,118 @@
...
+			width: '100%'
+		},
+		mobileBtnContainer: {
+			display: 'flex',
+			flexDirection: 'column',
+			padding: '3%'
+		}
+	};
+};
+function MenuBar({ classes }) {
```
TimothyGu, line 77: This file uses React Hooks, and thus should use the newer `useStyles` pattern rather than the `withStyle` pattern. Concretely, this should be a few changes:

1. Change the `const styles = (theme) => { ... };` to `const useStyles = makeStyles((theme) => { ... });`. See https://github.com/uclaacm/hoth.uclaacm.com/blob/a211aba2b32de50ed3bef48353feacee13b95bfa/src/components/HomePage/HothDescription.js#L13-L17

2. Apply the following change:
   ```suggestion
function MenuBar() {
  const classes = useStyles();
```

3. Remove `MenuBar.propTypes`.

4. Change `export default withStyles(styles)(MenuBar);` to just `export default MenuBar;`.

File: src/components/MenuBar/ButtonBar.js [hunk 8]
```diff
@@ -0,0 +1,70 @@
...
+				<Button fullWidth={isMobile} className={classes.btn} {...props} />
+            </LinkNoStyle>;
+    const BorderLink = ({ to, ...props }) =>
+        <LinkNoStyle>
+            <Button variant="outlined" fullWidth={isMobile} className={classes.btn} {...props} />
+        </LinkNoStyle>
+
+		const links = [
+			{
+				name: "home",
```
TimothyGu, line 30: These should be properly capitalized. Even though the CSS will make them appear all caps, we should keep the HTML semantically correct.

```suggestion
				name: "Home",
```

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,122 @@
...
+			</AppBar>
+
+			{/* Mobile menu Bar */}
+			<div className={classes.mobileMenuBar}>
+				<Collapse in={menuOpen}>
+					<div className={classes.mobileBtnContainer}>
+						<ButtonBar isMobile />
+					</div>
+				</Collapse>
+			</div>
```
TimothyGu, line 112: Currently, this is always rendered even if we are on desktop; it’s just that it’s hidden through CSS. It would be a cleaner (and speedier, but that’s less important) design if we could just not render them at all.

Check out https://material-ui.com/components/use-media-query/#using-material-uis-breakpoint-helpers for a way to detect the screen size using hooks.

Conversely, make the div labeled “Desktop menu Bar” only rendered on desktop.

```js
const isMobile = …; // check material-ui docs
const desktopMenuBar = isMobile ? null : <div>…</div>;
const mobileMenuBar = isMobile ? <div>…</div> : null;

return (
  <>
    …
    {desktopMenubar}
    …
    {mobileMenuBar}
  </>
);
```

src/components/MenuBar/ButtonBar.js
```diff
@@ -0,0 +1,89 @@
...
+
+	return (
+		<>
+			{links.map((link, index) =>
+				<PoppinLink to={link.to} key={`nav-${index}`}>
+					{link.name}
+				</PoppinLink>)}
+			<BorderLink to='/'>
+				Apply
+			</BorderLink>
```
TimothyGu, line 75: Maybe we would want a way to make this show up as “disabled” when we don’t have an active application going on right now? Maybe we could use Material-UI’s [outlined buttons](https://material-ui.com/components/buttons/#outlined-buttons) instead?

TimothyGu:
Also there is a merge conflict now. Make sure to run `yarn` after you merge `master` in.

src/components/MenuBar/ButtonBar.js
milesswu: We are already using the material-ui outlined buttons, I will update it so that the button is disabled for now

src/components/MenuBar/MenuBar.js
milesswu: Implemented in 'Navbar rendering' commit. Currently using a hardcoded 600 width value for the breakpoint. Cannot change this till a themeprovider is implemented.

TimothyGu: A ThemeProvider has been implemented on master.

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,148 @@
...
+	const classes = useStyles();
+	const theme = useTheme();
+	const [menuOpen, setMenuOpen] = useState(false);
+	const toggleMenu = () => setMenuOpen(open => !open);
+	const isMobile = useMediaQuery(theme.breakpoints.down(theme.breakpoints.values.sm * 1.3));
+
+	const desktopMenuBar =
+		<AppBar position="sticky">
+			<Toolbar className={classes.toolbar}>
+				<div
```
TimothyGu, line 95:
```suggestion
				<nav
```

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,148 @@
...
+		}
+	};
+});
+
+function MenuBar() {
+	const classes = useStyles();
+	const theme = useTheme();
+	const [menuOpen, setMenuOpen] = useState(false);
+	const toggleMenu = () => setMenuOpen(open => !open);
+	const isMobile = useMediaQuery(theme.breakpoints.down(theme.breakpoints.values.sm * 1.3));
```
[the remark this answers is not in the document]
jamieliu386: After removing some stuff as per [this comment](https://github.com/uclaacm/hoth.uclaacm.com/pull/15#discussion_r443243895), we no longer need the `menuBarAdaptiveThreshold` in `makeStyles` at all, so there's no more duplication.

src/components/MenuBar/MenuBar.js
jamieliu386: Should just the logo and wordmark be enclosed in `nav` tags? Or should the `ButtonBar` also be enclosed in `nav` tags rather than `div`?

src/components/MenuBar/MenuBar.js
```diff
@@ -0,0 +1,148 @@
...
+					<div
+						className={classes.logohome}
+						onClick={() => navigate('/')}
+					>
+						<HackLogo className={classes.logo} />
+						<Typography variant='h6' className={classes.wordmark} component='h6'>
+							HOTH
+						</Typography>
+					</div>
+					<IconButton onClick={toggleMenu} className={classes.menubtn}>
```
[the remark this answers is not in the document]
jamieliu386: Thanks for the suggestions!

Issue #14, opened 2020-05-23
Add Workshops Page


Timthetic:
Design: https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=15%3A9
Idea is to have 3 components:
* Schedule: Holds the entire schedule
* TimeSlot: Holds timestamp and events at that time
* Event: Holds event title, location, and description

PR #24, opened 2020-05-30
Add Workshop schedule page

Timthetic:
Will Fix: #14


src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+import Container from '@material-ui/core/Container';
+import Grid from '@material-ui/core/Grid';
+
+import { Typography } from '@material-ui/core';
```
TimothyGu, line 7: To be consistent with the other imports in this file, let's change it to the following style:
```suggestion
import Typography from '@material-ui/core/Typography';
```

For the differences, check https://material-ui.com/guides/minimizing-bundle-size/#development-environment.

src/components/header.js
```diff
@@ -27,6 +27,17 @@ const Header = ({ siteTitle }) =>
...
 			</h1>
+			<h2 style={{ margin: 0 }}>
+				<Link
+					to="/schedule"
+					style={{
+						color: `white`,
+						textDecoration: `none`
+					}}
+				>
+					Squedool
```
TimothyGu, line 38: This is fine, with the understanding that it’ll be obsolescent with #15.

src/pages/schedule.js
```diff
@@ -0,0 +1,21 @@
...
+import Event from '../components/SchedulePage/Event';
+
+const SchedulePage = () => {
+	const testEvents = [
+		{
+			eventName: 'ACM Hack Does a Back Flip',
+			eventLocation: 'Bolter Hall',
+			// eslint-disable-next-line no-multi-str
+			eventDescription: 'Tim Gu is very athletic and will demonstrate multiple back flipping techniques.\
+			Come by to learn about back flips and their history.'
```
TimothyGu, line 13: How about we use the events from last year to make sure this actually works?

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+Event.propTypes = {
+	classes: PropTypes.object.isRequired,
+	eventName: PropTypes.string.isRequired,
+	eventLocation: PropTypes.string.isRequired,
+	eventDescription: PropTypes.string.isRequired
+};
+
+export default withStyles(styles)(Event);
```
TimothyGu, line 51: Let's use the `makeStyles`/`useStyles` pattern found in https://github.com/uclaacm/hoth.uclaacm.com/blob/f218f58eb0ba969f345faed7bd149233860fed09/src/components/HomePage/HothDescription.js#L13-L23 for all new code.

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,51 @@
...
+		fontWeight: 'bold',
+		textTransform: 'uppercase'
+	},
+	eventLocation: {
+		fontFamily: theme.typography.fontFamily,
+		fontWeight: theme.typography.fontWeightMedium
+	}
+});
+
+function Event(props) {
```
TimothyGu, line 22: nit: we can destructure all the props in here.

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,51 @@
...
+			</Grid>
+		</Container>
+	);
+}
+
+Event.propTypes = {
+	classes: PropTypes.object.isRequired,
+	eventName: PropTypes.string.isRequired,
+	eventLocation: PropTypes.string.isRequired,
+	eventDescription: PropTypes.string.isRequired
```
TimothyGu, line 48:
```suggestion
	name: PropTypes.string.isRequired,
	location: PropTypes.string.isRequired,
	description: PropTypes.string.isRequired
```

We are in `Event`, so the `event` wouldn't be necessary :)

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,51 @@
...
+		fontFamily: theme.typography.fontFamily,
+		fontWeight: theme.typography.fontWeightMedium
+	}
+});
+
+function Event(props) {
+	return (
+		<Container maxWidth="md">
+			<Grid container spacing={1}>
+				<Grid item xs={12} sm={12} md={4}>
```
TimothyGu, line 26: I suspect that we could either leave off `xs={12}`, or just say `xs`, to make the `sm` value apply to `xs` automatically. Could you experiment a bit here and tell us what you find?

src/components/SchedulePage/Event.js
Timthetic: The other pattern here is in many other spots in the code base. Should we change those too? Otherwise, we will have two different patterns.

src/components/SchedulePage/Event.js
Timthetic: It looks like leaving off `xs={12}` works. It will use the value for given for `sm` when it passes the threshold for `xs`. Additionally, as a less relevant note, it seems that if you specify something like `xs={12} md={4}`, it will use the value for `xs` when going lower than the threshold for `md`. So I think even using that would work. I'm unsure what the convention is.

src/components/SchedulePage/Event.js
TimothyGu: Yeah. I'm thinking for now we can leave the copied code in the old style, but new code in the new style.

src/components/SchedulePage/Event.js
TimothyGu: It's up to you, though here are some of my thoughts.

To me, specifying `sm` and `md` would be more illuminating, as there is no question that `xs` would follow `sm`. On the other hand, specifying `xs` and `md` would leave me wondering what `sm` would be like: `xs` or `md`?

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,99 @@
...
+
+// Must be ordered
+const testEvents = [
+	{
+		name: 'ACM-W X ACM Hack | Hackathon 101 + Team Formation',
+		location: 'Sycamore Room',
+		// eslint-disable-next-line no-multi-str
+		description: 'First hackathon? We\'ll show you all the tips and tricks you\'ll ' +
+		'need to know to be successful and most importantly, have fun!',
+		startTime: '9:30 AM'
```
Timthetic, line 14: I didn't know whether to use a `Date` here. As of now, times act like an arbitrary group, and the events are clumped together similar to this: 'aabaaacc' => 'abac'

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,38 @@
...
+}));
+
+function Timeslot({ time, events }) {
+	const classes = useStyles();
+	return (
+		<Container maxWidth="md">
+			<Grid container spacing={1}>
+				<Grid item xs={12}>
+					<Divider />
+					<Typography className={classes.time}>{time}</Typography>
```
Timthetic, line 23: I'm unsure where to get the clock asset.

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,56 @@
...
+const useStyles = makeStyles(theme => ({
+	name: {
+		fontFamily: theme.typography.fontFamily,
+		color: theme.palette.secondary.main,
+		fontWeight: 'bold',
+		textTransform: 'uppercase'
+	},
+	location: {
+		fontFamily: theme.typography.fontFamily,
+		fontWeight: theme.typography.fontWeightMedium
```
jamieliu386, line 17: I was unaware that fontWeightMedium existed as a default theme value! Maybe we can consider overriding the defaults in `layout.js` and replacing hardcoded values elsewhere with them.

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,56 @@
...
+	},
+	description: {
+		fontFamily: theme.typography.fontFamily
```
jamieliu386, line 20: As a sidenote, I've noticed that most of our text is styled using Poppins in the figma, perhaps we can change the defaults in `layout.js` to Poppins and overriding in individual components when we want to use other fonts.

src/components/SchedulePage/Schedule.js
jamieliu386: I think storing the times as `Date` objects to might be a good idea to make things easier down the road with the upcoming workshops section. Maybe we can do something like this:
``` javascript
const hothDate = '2021-02-23'; // replace with real date later
const utcOffset = '-08:00'; // PST

const testEvents = [
  {
  	title: "ACM does a backflip",
  	time: new Date(hothDate + 'T09:30:00' + utcOffset)
  }
];
```

I'm also wondering if perhaps it'd be a good idea to have the data structure arranged such that each object has a time and a list of events starting at that time? e.g,
``` javascript
const testEvents = [
  {
  	time: new Date(hothDate + 'T10:00:00' + utcOffset),
    events: [
      {
        name: 'Git and Github',
        location: 'De Neve Auditorium',
        // eslint-disable-next-line no-multi-str
        description: 'With multiple developers each working on ' +
        'a different feature of a project, bringing things together ' +
        'can get messy pretty easily. Come learn how to use Git to ' +
        'manage code across multiple collaborators and versions!',
        startTime: '10:00 AM'
      },
      {
        name: 'INTRO TO IOS',
        location: 'Sycamore Room',
        // eslint-disable-next-line no-multi-str
        description: 'Learn all about the fundamentals of iOS development ' +
        'by making a simple app in Swift and XCode! ' +
        'You must have a Mac to be able to code along in this workshop.',
        startTime: '10:00 AM'
      },
    ]
  },
  ...
];
```
This would circumvent the need to create the `(time, [events]))` pairs later on. What do you think?

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,99 @@
...
+import Typography from '@material-ui/core/Typography';
+import TimeSlot from '../SchedulePage/Timeslot';
+
+// Must be ordered
+const testEvents = [
+	{
+		name: 'ACM-W X ACM Hack | Hackathon 101 + Team Formation',
+		location: 'Sycamore Room',
+		// eslint-disable-next-line no-multi-str
+		description: 'First hackathon? We\'ll show you all the tips and tricks you\'ll ' +
```
jamieliu386, line 12: We can replace the apostrophes with right single quotes.
```suggestion
		description: 'First hackathon? We’ll show you all the tips and tricks you’ll ' +
```

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,99 @@
...
+		startTime: '9:30 AM'
+	},
+	{
+		name: 'Git and Github',
+		location: 'De Neve Auditorium',
+		// eslint-disable-next-line no-multi-str
+		description: 'With multiple developers each working on ' +
+		'a different feature of a project, bringing things together ' +
+		'can get messy pretty easily. Come learn how to use Git to ' +
+		'manage code across multiple collaborators and versions!',
```
jamieliu386, line 23: I feel like we should avoid disabling eslint's no-multi-str, but I'm not sure what we would replace it with. 

I think we could instead use a React Fragment, i.e.
``` javascript
description: <>With multiple developers each working on
		a different feature of a project, bringing things together
		can get messy pretty easily. Come learn how to use Git to
		manage code across multiple collaborators and versions!</>
```
but I'd prefer for @TimothyGu to weigh in.

src/components/SchedulePage/Timeslot.js
jamieliu386: Material-UI has some good options if you search for 'clock': https://material-ui.com/components/material-icons/

src/components/SchedulePage/Schedule.js
Timthetic: Yeah, I was unsure about how we wanted to handle the data. But the way you are preposing is one I considered. The only think is that formatting dates seems to be a bit of a pain in JS. There are some packages that help, but idk if it's worth adding another package.

Also, I considered grouping things be start time as well. I didn't do that because then it's really more of a list of "time slots" than a list of "events" which feels like it breaks an abstraction somewhere. However, if you think that that's a better way, I can do it that way. I was on the fence about a lot of it.

– sent from my iPhone while laying in bed

src/components/SchedulePage/Schedule.js
Timthetic: Hmmm, I'll experiment with this. Seems clever. I didn't think to try this. I'll be back with results.

src/components/SchedulePage/Event.js
Timthetic: It looks like we have to default formatting comments. Should I fix this here, or might it be better to do it in a new PR?

src/components/SchedulePage/Schedule.js
Timthetic: This looks like it works. We can always undo it if they're issues.

TimothyGu: I agree with @jamieliu386 that we should avoid overriding ESLint whenever possible.

Generally as a rule, I'd rather not use JSX with text. There are some known pitfalls with doing so, like https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx.

How about using a template literal instead?
```js
		description: `With multiple developers each working on
		a different feature of a project, bringing things together
		can get messy pretty easily. Come learn how to use Git to
		manage code across multiple collaborators and versions!`,
```

TimothyGu:
@Timthetic In the future, can you click "Resolve conversation" for comments that you have addressed? This makes it clearer what the reviewer should look at.

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+import TimeSlot from '../SchedulePage/Timeslot';
+
+const hothDate = '2021-02-23'; // replace with real date later
+const utcOffset = '-08:00'; // PST
+const events = [
+	{
+		startTime: new Date(hothDate + 'T09:30:00' + utcOffset),
```
TimothyGu, line 10: Sorry, I know this has already been discussed, but I think it would generally be cleaner to use a function rather than duplicating the concatenation logic. Consider:

```suggestion
// time is a string of the form "HH:mm:ss"
function getDateObj(time) {
  return new Date(`${hothDate}T${time}${utcOffset}`);
}

const events = [
	{
		startTime: getDateObj('09:30:00'),
```

Additionally, you might even want to look into sharing the `hothDate` variable somehow with https://github.com/uclaacm/hoth.uclaacm.com/blob/581b54a8d3794ed0578370af2977a79a86c3e2b3/src/components/HomePage/Banner.js#L14

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+				description: <>First hackathon? We’ll show you all the tips and tricks you’ll
+				need to know to be successful and most importantly, have fun!</>
+			}
+		]
+	},
+	{
+		startTime: new Date(hothDate + 'T10:00:00' + utcOffset),
+		events: [
+			{
+				name: 'Git and Github',
```
TimothyGu, line 24:
```suggestion
				name: 'Git and GitHub',
```

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+				location: 'De Neve Auditorium',
+				description: <>With multiple developers each working on
+				a different feature of a project, bringing things together
+				can get messy pretty easily. Come learn how to use Git to
+				manage code across multiple collaborators and versions!</>
+			},
+			{
+				name: 'INTRO TO IOS',
```
TimothyGu, line 32: Always use normally-cased text for accessibility. If you want the text to show up as all-caps in the UI, set the `text-transform: uppercase` CSS property instead: https://github.com/uclaacm/hoth.uclaacm.com/blob/e9119ba6af7eba66eca17d2912c8a4975f7d09a0/src/components/Footer/Footer.js#L35

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+				You must have a Mac to be able to code along in this workshop.</>
+			}
+		]
+	},
+	{
+		startTime: new Date(hothDate + 'T11:00:00' + utcOffset),
+		events: [
+			{
+				name: 'FACEBOOK | ZERO TO HERO WITH REACT NATIVE',
+				location: 'De Never Auditorium ',
```
TimothyGu, line 45:
```suggestion
				location: 'De Neve Auditorium',
```

"De Never" sounds like a band name.

src/components/SchedulePage/Schedule.js
TimothyGu: (Of course, my recommendation only applies if you don't have to create additional elements.)

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+				description: <>In this workshop we will learn how to quickly create a mobile app
+				using React Native. We will be covering the React fundamentals (such as props
+				and state) and apply them to quickly build a simple cross-platform app using
+				React Native. This will be a hands-on workshop, so come in with the intent to hack!
+				<br/>If possible, come in with Node.js, Git, and a text editor of your choice installed already!</>
```
TimothyGu, line 50: Use separate `<p>` elements rather than `<br>`.

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+		fontWeight: 'bold',
+		paddingTop: 24,
+		paddingBottom: 16
+	}
+}));
+
+const SchedulePage = () => {
+	const classes = useStyles();
+	return <>
+		<Typography align='center' variant='h4' className={classes.title}>Schedule</Typography>
```
TimothyGu, line 82:
```suggestion
		<Typography align='center' component='h1' variant='h4' className={classes.title}>Schedule</Typography>
```

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,58 @@
...
+const formatTime = time => {
+	let hours = time.getHours();
+	let minutes = time.getMinutes();
+	const ampm = hours >= 12 ? 'PM' : 'AM';
+	hours %= 12;
+	hours = hours ? hours : 12; // the hour '0' should be '12'
+	minutes = minutes < 10 ? '0' + minutes : minutes;
+	const strTime = hours + ':' + minutes + ' ' + ampm;
+	return strTime;
+};
```
TimothyGu, line 31: 1. Change `time` to `date`, as the input is a `Date` object.

2. Even better, use [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat).

   ```js
   const timeFormatter = new Intl.DateTimeFormat('en-US', { timeStyle: 'short' });
   // ...
   <Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
   ```

src/components/SchedulePage/Schedule.js
Timthetic: @TimothyGu We can't have line breaks then, but that seems like a sacrifice that I'd be okay with making.

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,58 @@
...
+
+function Timeslot({ time, events }) {
+	const classes = useStyles();
+	return (
+		<Container maxWidth="md">
+			<Grid container spacing={1}>
+				<Grid item xs={12}>
```
TimothyGu, line 38: I think it would be best if you find a way to avoid the use of `Grid`. Grid is designed to be primarily used for horizontal layout, like if we have several elements side-by-side. I have a suspicion that simply removing the two `Grid` elements would work just fine.

Additionally, the `Container` part should go in the `Schedule` component, surrounding both the title and this class. A rule of thumb is that you generally don't need have a `Container` inside another `Container`. If you want the `<Event>` to have a left margin, add one manually with CSS rather than relying on `Container`, which can contribute to inconsistencies like the following.

Desktop:

image

Mobile:

image

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,56 @@
...
+	},
+	container: {
+		paddingBottom: 12
+	}
+}));
+
+function Event({ name, location, description }) {
+	const classes = useStyles();
+	return (
+		<Container maxWidth="md">
```
TimothyGu, line 30: See my comment in `Timeslot.js`. I don't think the `Container` should be needed.

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,89 @@
...
+import TimeSlot from '../SchedulePage/Timeslot';
+
+const hothDate = '2021-02-23'; // replace with real date later
+const utcOffset = '-08:00'; // PST
+const events = [
+	{
+		startTime: new Date(hothDate + 'T09:30:00' + utcOffset),
+		events: [
+			{
+				name: 'ACM-W X ACM Hack | Hackathon 101 + Team Formation',
```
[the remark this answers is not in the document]
Timthetic: I will commit this, but all of this will likely be overwritten with new workshops.

src/components/SchedulePage/Schedule.js
Timthetic: I like both of these suggestions. Thanks @TimothyGu

src/components/SchedulePage/Schedule.js
TimothyGu: Feel free to override this if you need to create additional elements. Also, see my other comment: we probably should not be using `<br>` anyway.

src/components/SchedulePage/Schedule.js
Timthetic: Also, @TimothyGu should we just add another file for HoTH "metadata" or something? (probably in another PR?)

src/components/SchedulePage/Schedule.js
Timthetic: For now, with template strings, I'm just going to remove the new line. We can come back to it later if it is necessary.

src/components/SchedulePage/Timeslot.js
Timthetic: @TimothyGu I'm confused how I would get them next to each other when the screen is large if we remove the Grid.

TimothyGu: I'm not asking you to remove the Grid in Event, where I assume it's necessary as you said to get the event name and description in horizontal layout. I'm saying that you probably don't need it in this file, Timeslot.js.

TimothyGu: To summarize:

In `Timeslot.js`, delete `Container` and `Grid`.

In `Event.js`, delete `Container`.

In `Schedule.js`, add `Container`.

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,52 @@
...
+import Divider from '@material-ui/core/Divider';
+import Event from '../SchedulePage/Event';
+import ScheduleRoundedIcon from '@material-ui/icons/ScheduleRounded';
+
+const useStyles = makeStyles(theme => ({
+	time: {
+		fontFamily: theme.typography.fontFamily,
+		marginLeft: 4
+	},
+	timeContainer: {
```
Timthetic, line 14: Here's a good place so look for review. I'm not great with CSS, so I'm unsure if my styling is good here.

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,95 @@
...
+		fontWeight: 'bold',
+		paddingTop: 24,
+		paddingBottom: 16
+	}
+}));
+
+const SchedulePage = () => {
+	const classes = useStyles();
+	return <>
+		<Typography align='center' component='h1' variant='h4' className={classes.title}>Schedule</Typography>
```
jamieliu386, line 87: The wordmark on the MenuBar is an `h1` component, so I think this should be an `h2`.
```suggestion
		<Typography align='center' component='h2' variant='h4' className={classes.title}>Schedule</Typography>
```

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,52 @@
...
+
+const timeFormatter = new Intl.DateTimeFormat('en-US', { timeStyle: 'short', hour: 'numeric', minute: '2-digit' });
+
+function Timeslot({ time, events }) {
+	const classes = useStyles();
+	return (
+		<div className={classes.timeSlot}>
+			<Divider />
+			<div className={classes.timeContainer}>
+				<ScheduleRoundedIcon fontSize={'inherit'} className={classes.clockImage}/>
```
jamieliu386: Looks like the `clockImage` class doesn't exist, so we can remove the `className` here.
```suggestion
				<ScheduleRoundedIcon fontSize={'inherit'} />
```

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,52 @@
...
+				<Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
+			</div>
+			{events.map(event => {
+				return <Event {...event} key={event.name}/>;
```
jamieliu386:
```suggestion
				return <Event {...event} key={event.name} />;
```

src/components/SchedulePage/Event.js
jamieliu386: > default formatting comments

what do you mean by that?

src/components/SchedulePage/Timeslot.js
jamieliu386: This looks fine to me, but I'm also not the best with CSS.

src/components/SchedulePage/Timeslot.js
```diff
@@ -34,12 +37,14 @@ function Timeslot({ time, events }) {
 		<div className={classes.timeSlot}>
 			<Divider />
 			<div className={classes.timeContainer}>
-				<ScheduleRoundedIcon fontSize={'inherit'} className={classes.clockImage}/>
-				<Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
+				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage}/>
+				<div>
+					<Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
+					{events.map(event => {
+						return <Event {...event} key={event.name}/>;
```
jamieliu386, line 44:
```suggestion
						return <Event {...event} key={event.name} />;
```

src/components/SchedulePage/Timeslot.js
```diff
@@ -34,12 +37,14 @@ function Timeslot({ time, events }) {
 		<div className={classes.timeSlot}>
 			<Divider />
 			<div className={classes.timeContainer}>
-				<ScheduleRoundedIcon fontSize={'inherit'} className={classes.clockImage}/>
-				<Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
+				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage}/>
```
jamieliu386, line 40:
```suggestion
				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage} />
```

src/components/SchedulePage/Schedule.js
```diff
@@ -0,0 +1,95 @@
...
+		fontFamily: theme.typography.fontFamily,
+		fontWeight: 'bold',
+		paddingTop: 24,
+		paddingBottom: 16
+	}
+}));
+
+const SchedulePage = () => {
+	const classes = useStyles();
+	return <>
```
TimothyGu, line 86: From my last round of reviews, I told you to add a `Container` in `Schedule.js`. See https://github.com/uclaacm/hoth.uclaacm.com/pull/24#discussion_r444615333.

Concretely, you can do so by just changing these `<>` to `<Container>`, and of course also `</>` to `</Container>`.

src/pages/schedule.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import Layout from '../components/layout';
+import Schedule from '../components/SchedulePage/Schedule';
+
+const SchedulePage = () => {
+	return <Layout>
+		<Schedule/>
```
TimothyGu, line 8: You should add a `<SEO />` element, like https://github.com/uclaacm/hoth.uclaacm.com/blob/e4e5b087ff8cf58aa9b222778b33121c6501cf3a/src/pages/index.js#L12.

src/components/SchedulePage/Timeslot.js
TimothyGu: Instead of using flexboxes directly, can you instead use the [`<Box>`](https://material-ui.com/system/flexbox/#flexbox) component?

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,57 @@
...
+		display: 'flex'
+	},
+	timeSlot: {
+		paddingLeft: '16px',
+		paddingRight: '16px',
+		[theme.breakpoints.up('md')]: {
+			width: theme.breakpoints.values.md,
+			marginLeft: 'auto',
+			marginRight: 'auto'
+		}
```
TimothyGu, line 28: There should be no need for this class if you had added a `<Container>` to `Schedule.js`, as I previously suggested.

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,57 @@
...
+
+function Timeslot({ time, events }) {
+	const classes = useStyles();
+	return (
+		<div className={classes.timeSlot}>
+			<Divider />
+			<div className={classes.timeContainer}>
+				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage}/>
+				<div>
+					<Typography className={classes.time}>{timeFormatter.format(time)}</Typography>
```
TimothyGu, line 42:
```suggestion
					<Typography component='h2' className={classes.time}>
						<time dateTime={time.toISOString()}>{timeFormatter.format(time)}</time>
					</Typography>
```

* Use `h2` to signify this is the secondary-level heading, after the page title "Schedule".
* Use `<time>` for semantically correct HTML element, accompanied by the datetime in ISO standard format. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time.

src/components/SchedulePage/Event.js
```diff
@@ -0,0 +1,53 @@
...
+		paddingBottom: 12
+	}
+}));
+
+function Event({ name, location, description }) {
+	const classes = useStyles();
+	return (
+		<Grid container spacing={1} className={classes.container}>
+			<Grid item sm={12} md={4}>
+				<Typography variant='subtitle1' className={classes.name}>
```
TimothyGu, line 31:
```suggestion
				<Typography component='h3' variant='subtitle1' className={classes.name}>
```
Use `<h3>` to signify this is the tertiary-level heading, after the time slot.

src/components/SchedulePage/Timeslot.js
```diff
@@ -0,0 +1,57 @@
...
+			width: theme.breakpoints.values.md,
+			marginLeft: 'auto',
+			marginRight: 'auto'
+		}
+	}
+}));
+
+const timeFormatter = new Intl.DateTimeFormat('en-US', { timeStyle: 'short', hour: 'numeric', minute: '2-digit' });
+
+function Timeslot({ time, events }) {
```
TimothyGu, line 34: Please be consistent with the capitalization. This component is referred to as `TimeSlot` elsewhere, which is proper CamelCase. Remember to rename the file as well.

src/components/SchedulePage/Event.js
Timthetic: Idk how I didn't catch this. What I wanted to say was "**two** default formatting comments suggesting to put something in `layout.js`".

jamieliu386: oh yeah, I think it's fine for us to do that in a separate PR since it'll involve cleaning up other unrelated files as well

src/pages/schedule.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Layout from '../components/layout';
+import SEO from '../components/seo';
+import Schedule from '../components/SchedulePage/Schedule';
+
+
+const SchedulePage = () => {
+	return <Layout>
+		<SEO title='Schedule' />
```
TimothyGu, line 10: Optional:

```suggestion
		<SEO title='Workshop schedule' />
```

TimothyGu:
Let's wait so that @jamieliu386 can take another look.

src/pages/schedule.js
jamieliu386: Will there be a separate overarching schedule eventually? Or will we include things like the opening ceremony, lunch, etc. on this page?

jamieliu386:
LGTM! Nice work :)

Issue #13, opened 2020-05-23
Create FAQ section


jamieliu386:
Create FAQ section as shown in the [figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=15%3A9)

TimothyGu:
We still need to tweak the styling after #18.

jamieliu386:
Discovered an issue where, when viewing from an iPhone 5/SE, there is a visual glitch when opening/closing questions that span multiple lines.

PR #31, opened 2020-06-15
Fix animation issue with iPhone 5/SE

jamieliu386:
Fixes: #13

### DIFF
--- a/src/components/HomePage/FAQSection.js
+++ b/src/components/HomePage/FAQSection.js
@@ -4,7 +4,7 @@ import { Typography, Link } from '@material-ui/core';
 import Container from '@material-ui/core/Container';
 import MuiExpansionPanel from '@material-ui/core/ExpansionPanel';
 import MuiExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary';
-import ExpansionPanelDetails from '@material-ui/core/ExpansionPanelDetails';
+import MuiExpansionPanelDetails from '@material-ui/core/ExpansionPanelDetails';
 
 import Add from '@material-ui/icons/Add';
 import Remove from '@material-ui/icons/Remove';
@@ -30,14 +30,15 @@ const ExpansionPanel = withStyles(theme => ({
 
 const ExpansionPanelSummary = withStyles({
 	root: {
-		minHeight: 20,
+		minHeight: 'auto',
 		'&$expanded': {
-			minHeight: 20
+			minHeight: 'auto'
 		}
 	},
 	content: {
+		margin: 5,
 		'&$expanded': {
-			margin: 0
+			margin: 5
 		}
 	},
 	expanded: {},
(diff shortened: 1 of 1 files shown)


src/components/HomePage/FAQSection.js
```diff
@@ -4,49 +4,58 @@ import { Typography, Link } from '@material-ui/core';
...
 import Add from '@material-ui/icons/Add';
 import Remove from '@material-ui/icons/Remove';
 
 const ExpansionPanel = withStyles(theme => ({
 	root: {
 		backgroundColor: theme.palette.background.default,
-		borderTop: '1px solid rgba(0, 0, 0, 1)',
-		borderBottom: '1px solid rgba(0, 0, 0, 1)',
+		borderTop: '1px solid black',
+		borderBottom: '1px solid black',
```
TimothyGu, line 16: Black is the same thing as the above, only better known.

src/components/HomePage/FAQSection.js
```diff
@@ -4,49 +4,58 @@ import { Typography, Link } from '@material-ui/core';
...
 		boxShadow: 'none',
 		'&:not(:last-child)': {
 			borderBottom: 0
 		},
-		'&:before': {
+		'&::before': {
```
TimothyGu, line 21: `::before` is a [pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) which ordinarily are prefixed by `::`. `:before` is a compatibility syntax, as a single `:` refers to [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/pseudo-classes).

src/components/HomePage/FAQSection.js
```diff
@@ -4,49 +4,58 @@ import { Typography, Link } from '@material-ui/core';
...
 			display: 'none'
 		},
 		'&$expanded': {
-			margin: 'auto'
+			margin: 0
```
TimothyGu, line 25: In this context, both are equivalent – but 0 makes a bit more sense.

src/components/HomePage/FAQSection.js
```diff
@@ -4,49 +4,58 @@ import { Typography, Link } from '@material-ui/core';
...
 	root: {
-		minHeight: 20,
+		minHeight: 'auto',
 		'&$expanded': {
-			minHeight: 20
+			minHeight: 'auto'
+		},
+		[theme.breakpoints.down('sm')]: {
+			paddingLeft: 8,
+			paddingRight: 8
```
TimothyGu, line 39: As discussed, this makes the page take up more space on the smallest devices that lack much display real estate. `ExpansionPanelDetails` is modified to match this.

src/components/HomePage/FAQSection.js
```diff
@@ -4,49 +4,58 @@ import { Typography, Link } from '@material-ui/core';
...
-			minHeight: 20
+			minHeight: 'auto'
+		},
+		[theme.breakpoints.down('xs')]: {
+			paddingLeft: 8,
+			paddingRight: 8
 		}
 	},
 	content: {
+		margin: '12px 0',
```
TimothyGu, line 43: 12px is chosen as that's what the framework does internally. 0 means no horizontal margin, which is also what the framework does (and which is all we need).

src/components/HomePage/FAQSection.js
jamieliu386: Oops, thanks for pointing this out!

jamieliu386:
Thanks for making these edits, Tim!

Issue #30, opened 2020-06-14
Various search engine optimization (SEO)–related matters


TimothyGu:
Currently, there are a few things that would be needed for a polished website that we do not yet have.

- [ ] `<title>` tag. Right now we don't have one, and the URL gets put in the tab heading instead. ![image](https://user-images.githubusercontent.com/1538624/84605445-40326680-ae6b-11ea-803c-3a40e418ed59.png)
- [ ] the `lang` attribute of `<html>`, which is important for accessibility
- [ ] other site metadata (like `<meta>` tags) that hack.uclaacm.com has

PR #40, opened 2020-06-20
Add SEO-related items

TimothyGu:
This PR adds the following to every page (admittedly only one at the moment):

- HTML `<title>`
- `lang` attribute of `<html>`
- `<meta>` tags (including [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name), [Open Graph](https://ogp.me/), and [Twitter](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup))
- [Web app manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest)

Also remove some stale stuff from the Gatsby boilerplate we started out with.

Fixes: #30


gatsby-config.js
```diff
@@ -17,15 +21,17 @@ module.exports = {
...
+				short_name: `HOTH`,
+				start_url: `/`,
+				background_color: `#fff`,
+				theme_color: `#c960ff`,
 				display: `minimal-ui`,
 				icon: `src/images/gatsby-icon.png` // This path is relative to the root of the site.
 			}
+			/* eslint-enable camelcase */
 		},
 		// this (optional) plugin enables Progressive Web App + Offline functionality
```
jamieliu386, line 36: Should 'gatsby-plugin-acmhack-svgr' be quoted using backticks?

TimothyGu: If I had a choice I would change all these backticks to quotation marks, as the template literals are a remnant of the Gatsby boilerplate. Maybe I’ll do that as a separate commit.

TimothyGu: Done.

Issue #46, opened 2020-06-23
Homepage countdown may be clipped on mobile


Timthetic:
<img width="380" alt="Screen Shot 2020-06-23 at 12 56 04 AM" src="https://user-images.githubusercontent.com/33593205/85376336-68196e00-b4ec-11ea-8b14-b14f245d20b0.png">
When the screen gets small enough, the countdown on the homepage exits the screen. Above is the homepage for the dimensions of "iPhone 6/7/8 plus"

jamieliu386:
Looks like the "Hack on the Hill 7" alignment is also off for small screens.

Thanks for catching this!

PR #47, opened 2020-06-23
Banner fixes

jamieliu386:
Fixes: #46 

Makes the countdown's font smaller on `xs` screens. Also fixes the text alignment for "Hack on the Hill 7" and fixes the date used for the countdown.

image

### DIFF
--- a/src/components/HomePage/Banner.js
+++ b/src/components/HomePage/Banner.js
@@ -24,7 +24,10 @@ const useStyles = makeStyles(theme => ({
 			backgroundRepeat: 'no-repeat',
 			backgroundPosition: 'right bottom'
 		},
-		padding: 25
+		padding: 25,
+		[theme.breakpoints.down('sm')]: {
+			padding: '12px 2px'
+		}
 	},
 	text: {
 		color: 'white',
@@ -42,17 +45,16 @@ const useStyles = makeStyles(theme => ({
 	timer: {
 		color: 'white',
 		fontWeight: 600,
-		fontSize: '4em',
-		textAlign: 'center',
+		fontSize: '3em',
 		[theme.breakpoints.down('xs')]: {
-			fontSize: '2em'
+			fontSize: '2.3rem'
 		}
 	},
 	timeDesc: {
 		'&::after': {
 			content: 'attr(data-field)',
 			fontWeight: 400,
-			fontSize: '0.23em'
+			fontSize: '0.9rem'
 		}
 	}
 }));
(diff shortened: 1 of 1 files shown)


src/components/HomePage/Banner.js
```diff
@@ -143,13 +149,15 @@ function Banner() {
 		completed: PropTypes.bool.isRequired
 	};
 
+	const smallScreen = useMediaQuery(theme.breakpoints.down('sm'));
+
 	return (
 		<div className={classes.container}>
```
TimothyGu, line 155: Idea: reduce the padding on the `.container` for smaller screen sizes. I think `padding: '12px 4px'` looks good.

src/components/HomePage/Banner.js
```diff
@@ -40,7 +42,11 @@ const useStyles = makeStyles(theme => ({
 	timer: {
 		color: 'white',
 		fontWeight: 600,
-		fontSize: '4em'
+		fontSize: '4em',
+		textAlign: 'center',
+		[theme.breakpoints.down('xs')]: {
```
TimothyGu, line 47: I think the addition of the rule

```js
[theme.breakpoints.down('sm')]: {
  fontSize: '3em'
}
```

would work even better than this change alone. Right now at the `sm` size, the timer still appears bigger than the "Hack on the Hill 7" text itself, which looks a bit strange.

src/components/HomePage/Banner.js
```diff
@@ -40,7 +42,11 @@ const useStyles = makeStyles(theme => ({
 	timer: {
 		color: 'white',
 		fontWeight: 600,
-		fontSize: '4em'
+		fontSize: '4em',
+		textAlign: 'center',
```
TimothyGu, line 46: The `textAlign` change doesn't appear necessary here, and I suspect that the `textAlign` change above might not be necessary either. I think the `alignItems` fix below might have rendered these unnecessary.

jamieliu386:
> While it does admit extra space between the colon and the number on small screens, I think it superior to making the units not readable at all.

How about making the font size bigger but also changing the units for small screens? i.e. dd : hh : mm : ss or something similar? I really don't like the look of the extra space between the number and colon.

src/components/HomePage/Banner.js
```diff
@@ -103,7 +106,7 @@ function Banner() {
 					>
 						{days.toString().padStart(2, '0')}
 					</Box>
-					:
+					<div style={{ padding: '0px 5px' }}>:</div>
```
TimothyGu, line 108: Can you move these styles into a class, like `className={classes.colon}`?

TimothyGu: I've taken care of it.

jamieliu386:
Thanks for the fixes, looks good!

Issue #38, opened 2020-06-20
PR labels and labeler


TimothyGu:
I've recently added some issue/PR labels for each component: documentation, infra, homepage, schedule, prizes, etc. This makes it easier for us to track the progress on each page.

However, labeling everything by hand could be a bit tedious. We could consider using the [labeler](https://github.com/actions/labeler) GitHub Actions, which will label PRs by the files changed.

TimothyGu:
This is not working yet.

PR #50, opened 2020-06-25
Add files for PR labeler

Timthetic:
Fixes: #38


.github/labeler.yml
```diff
@@ -0,0 +1,17 @@
...
+- any: ['src/components/PrizePage/**/*', 'src/pages/prizes.js']
+
+page:
+- any: ['src/components/MenuBar/**/*', 'src/components/Footer/**/*', 'src/components/layout.js']
+
+gallery:
+- any: ['src/components/GalleryPage/**/*', 'src/pages/gallery.js']
+
+infra:
+- any: ['.github/**/*', '!.github/FUNDING.yml']
```
TimothyGu, line 17: Let's also add these to `infra`
```
src/components/SvgImg.js
.editorconfig
.eslint*
.huskyrc.js
.nvmrc
.prettier*
.svgo*
gatsby-*.js
package.json
yarn.lock
```

And the following to `documentation`:
```
LICENSE*
*.md
.github/FUNDING.yml
```

Finally, make sure you add a new line to the end of files.

.github/labeler.yml
```diff
@@ -0,0 +1,17 @@
...
+- any: ['src/components/SchedulePage/**/*', 'src/pages/schedule.js']
+
+homepage:
+- any: ['src/components/HomePage/**/*', 'src/pages/index.js']
+
+prizes:
+- any: ['src/components/PrizePage/**/*', 'src/pages/prizes.js']
+
+page:
+- any: ['src/components/MenuBar/**/*', 'src/components/Footer/**/*', 'src/components/layout.js']
```
TimothyGu, line 11: Also `src/components/seo.js`

.github/labeler.yml
TimothyGu, line 17: Another thing is that I think you meant to say "anything in `.github`, except `FUNDING.yml`". If this is the case, then you should do

```yaml
infra:
- any: ['.github/**/*']
  all: ['!.github/FUNDING.yml']
```

See https://github.com/actions/labeler#common-examples

.github/labeler.yml
```diff
@@ -0,0 +1,17 @@
+schedule:
+- any: ['src/components/SchedulePage/**/*', 'src/pages/schedule.js']
```
TimothyGu, line 2: This probably doesn't do what you want it to do. https://github.com/actions/labeler#match-object says that `any` means

> match ALL globs against ANY changed path.

It's obviously impossible for any changed path to match both of these globs.

Instead, you should do this:

```yaml
schedule:
- src/components/SchedulePage/**/*
- src/pages/schedule.js
```

Same goes for everything else.

.github/labeler.yml
Timthetic: Wait, wouldn't that one be: "any change to `.github/**/*` files as long as the `.github/FUNDING.yml` hasn't changed"

TimothyGu: Ah, I think you're right. The status quo looks good then.

Issue #42, opened 2020-06-21
Upcoming events on the homepage


TimothyGu:
See [Figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=0%3A1)

Timthetic:
I thought this would be trivial, because I could just reuse the `TimeSlot` component. It seems that my assumption was incorrect. I'm unsure if I should try to adapt the `TimeSlot` component to work here or to create a new one.

<img width="954" alt="HoTH-dev-Upcoming" src="https://user-images.githubusercontent.com/33593205/85900226-4fdb7480-b7b4-11ea-9628-1f28c6813d75.png">

TimothyGu:
Please submit a PR first so we can discuss!

PR #52, opened 2020-06-27
Upcoming workshops

Timthetic:
Fixes: #42


src/components/HomePage/UpcomingEvents.js
```diff
@@ -0,0 +1,45 @@
...
+	const nextTimeSlot = workshopSchedule.reduce((closestTimeSlot, timeSlot) => {
+		if ((currentTime < timeSlot.startTime && timeSlot.startTime < closestTimeSlot.startTime) ||
+			currentTime > closestTimeSlot.startTime) {
+			return timeSlot;
+		}
+		return closestTimeSlot;
+	}, workshopSchedule[0]);
+	return <Container maxWidth='md' className={classes.eventsContainer}>
+		<Grid container justify='center' spacing={1}>
+			<Grid item md={11}>
```
TimothyGu, line 35: Ah, we really should try to avoid this. I believe the reason why it's necessary is because of the fact that HothDescription is split into 4/7, which doesn't add up to 12. If you change that component to be 4/8, then it'll look better.

src/data/WorkshopSchedule.js
```diff
@@ -0,0 +1,70 @@
...
+				location: 'De Neve Learning Center',
+				description: `Learn all about the fundamentals of iOS development
+				by making a simple app in Swift and XCode!
+				You must have a Mac to be able to code along in this workshop.`
+			}
+		]
+	}
+];
+
+module.exports = workshopSchedule;
```
TimothyGu, line 70:
```suggestion
export default workshopSchedule;
```

src/components/HomePage/UpcomingEvents.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
```
TimothyGu, line 2:
```suggestion
```

src/components/SchedulePage/TimeSlot.js
```diff
@@ -26,20 +22,17 @@ const timeFormatter = new Intl.DateTimeFormat('en-US', { timeStyle: 'short', hou
...
-						<time dateTime={time.toISOString()}>{timeFormatter.format(time)}</time>
-					</Typography>
-					{events.map(event => {
-						return <Event {...event} key={event.name} />;
-					})}
-				</div>
-			</Box>
-		</>
+		<Box display='flex'>
+			<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage} />
```
TimothyGu, line 26:
```suggestion
			<ScheduleRoundedIcon fontSize='small' className={classes.clockImage} />
```

Issue #53, opened 2020-06-29
Change Default Fonts to Poppins


Timthetic:
Change the `overwrittenTheme` in `layout.js` to use Poppins as the default font for everything.

PR #54, opened 2020-06-29
Styling Fixes

Timthetic:
Fixes: #53

### DIFF
--- a/src/components/SchedulePage/Event.js
+++ b/src/components/SchedulePage/Event.js
@@ -26,7 +26,7 @@ function Event({ name, location, description }) {
 				<Typography component='h3' variant='subtitle1' className={classes.name}>
 					{name}
 				</Typography>
-				<Typography variant='body1' className={classes.location}>
+				<Typography variant='body2' className={classes.location}>
 					{location}
 				</Typography>
 			</Grid>
--- a/src/components/SchedulePage/Event.js
+++ b/src/components/SchedulePage/Event.js
@@ -6,17 +6,12 @@ import Typography from '@material-ui/core/Typography';
 
 const useStyles = makeStyles(theme => ({
 	name: {
-		fontFamily: theme.typography.fontFamily,
 		color: theme.palette.secondary.main,
 		fontWeight: 'bold',
 		textTransform: 'uppercase'
 	},
 	location: {
-		fontFamily: theme.typography.fontFamily,
 		fontWeight: theme.typography.fontWeightMedium
-	},
-	description: {
-		fontFamily: theme.typography.fontFamily
 	},
 	container: {
 		paddingBottom: 12
@@ -36,7 +31,7 @@ function Event({ name, location, description }) {
 				</Typography>
 			</Grid>
 			<Grid item sm={12} md={8}>
-				<Typography variant='body1' className={classes.description}>
+				<Typography variant='body1'>
 					{description}
 				</Typography>
 			</Grid>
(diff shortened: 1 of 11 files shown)


TimothyGu:
Just my opinion, so take it as a grain of salt. But I personally prefer Chivo at smaller font sizes and longer texts. I agree that we can probably do away with Palanquin though.

Another issue that's more blocking is please remember to revise the list of imported fonts, here:

https://github.com/uclaacm/hoth.uclaacm.com/blob/becb3a6b71e45256e32f1ba54aaa605751281ef3/src/components/layout.js#L52-L53

jamieliu386:
How about we edit the figma to use Chivo for paragraphs or blocks of smaller text? 

I'd like the website to reflect the design in the figma, which currently is almost completely Poppins.

After we fix the figma, we can edit this PR to include those changes as well.

Timthetic:
> How about we edit the figma to use Chivo for paragraphs or blocks of smaller text?
> I'd like the website to reflect the design in the figma, which currently is almost completely Poppins.
> After we fix the figma, we can edit this PR to include those changes as well.
> 
> Edit: I changed some of the fonts in the figma. Comparing this preview with the current live site, I agree that Chivo is nicer for the readability of blocks of text. I think it's a good idea to add the `body1` override back into the theme, and we may also want to edit the schedule page to have the workshop descriptions have `variant='body1'`, too.

I agree that the Figma should be consistent with our design. If we want to use Chivo, that's perfectly fine. I'm down for changing `body1` to use Chivo by default.

src/components/layout.js
```diff
@@ -32,13 +32,6 @@ const overwrittenTheme = responsiveFontSizes(createMuiTheme({
...
 		fontFamily: '"Poppins", sans-serif',
-		h1: { fontFamily: '"Palanquin Dark", sans-serif' },
-		h2: { fontFamily: '"Palanquin Dark", sans-serif' },
-		h3: { fontFamily: '"Palanquin Dark", sans-serif' },
-		h4: { fontFamily: '"Palanquin Dark", sans-serif' },
-		h5: { fontFamily: '"Palanquin Dark", sans-serif' },
-		h6: { fontFamily: '"Palanquin Dark", sans-serif' },
-		subtitle1: { fontFamily: '"Palanquin Dark", sans-serif' },
 		body1: { fontFamily: '"Chivo", sans-serif' },
 		button: { fontFamily: '"Chivo", sans-serif' }
```
jamieliu386, line 36: I think we can remove the button override
```suggestion
```

src/components/SchedulePage/Event.js
```diff
@@ -36,7 +31,7 @@ function Event({ name, location, description }) {
 				</Typography>
```
jamieliu386, line 31: Two lines up, let's change the location to not have `variant='body1'`, since I think it should be using Poppins.

Likewise, we need to remove the `variant='body1'` attribute from line 134 of the `FAQSection`, since each of the questions should be in Poppins.

src/components/layout.js
TimothyGu: This leads to the navbar buttons to be rendered as Chivo:

image

I believe they should be Poppins, as suggested by @jamieliu386.

src/components/SchedulePage/Event.js
TimothyGu: My recommendation is to use `variant='body2'` for the location. It's still a kind of body, except it uses Poppins.

Be mindful, though, that when you remove `variant='body1'` or replace it with `body2`, the font size gets smaller. For location, this is probably fine; but for the FAQ headings, I believe the font size should stay at its current size.

src/components/layout.js
Timthetic: I actually like the top as well now that I look at it side by side

Issue #55, opened 2020-07-01
Netlify changes


TimothyGu:
A couple of changes we probably want to make to our Netlify deployment.

* Use [file-based configuration](https://docs.netlify.com/configure-builds/file-based-configuration/). Right now, there are some configurations (build command and build output directory, and perhaps some more) that are set through the Netlify UI itself. We should store those settings inside this repo, by creating a `netlify.toml` file.

* Netlify runs on UTC time by default, which leads to incorrectly rendered time in the built HTML file. (The reason why it's not a big issue right now is since the page is re-rendered dynamically in React, but you can see the issue if you disable JavaScript.) According to https://community.netlify.com/t/can-i-set-a-custom-timezone-in-the-build-environment/1915, fixing this should be as simple as

   ```toml
   [build.environment]
     TZ = "America/Los_Angeles"
   ```

   in the `netlify.toml` file.

PR #56, opened 2020-07-03
Add Netlify configuration file

TimothyGu:
Fixes: #55

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,8 @@
+# https://docs.netlify.com/configure-builds/file-based-configuration/
+
+[build]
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -0,0 +1,8 @@
+# https://docs.netlify.com/configure-builds/file-based-configuration/
+
+[build]
+publish = "public/"
+command = "yarn build"
```
jamieliu386, line 5: Should these be indented? (likewise for line 8)

TimothyGu: As discussed, I'd rather not indent.

Issue #57, opened 2020-07-03
Time zone–proofing the website


TimothyGu:
We need to prepare for the eventuality that HOTH may be hosted (at least in part) remotely. This means we have to do, what I'll call time zone–proofing this website.

First, we have to decide which time zone to use. It would be easier if we just stick with America/Los\_Angeles for everything, but it's probably not the most friendly for folks outside of the West Coast.

If we were to decide on local time zones, we have to do a few things:

- Make it clear that the times are local, perhaps by displaying it. Most people may just assume times are LA time. One can get the time zone name this way:

   ```js
   // for short names, like EST or GMT+5:30
   new Intl.DateTimeFormat('en-US', { timeZoneName: 'short' }).formatToParts(new Date('2020-01-01')).slice(-1)[0].value
   // for long names, like Eastern Standard Time or India Standard Time
   new Intl.DateTimeFormat('en-US', { timeZoneName: 'long' }).formatToParts(new Date('2020-01-01')).slice(-1)[0].value
   ```

- Detect if the event crosses the day boundary, and change the banner as appropriate. (Like February 29<sup>th</sup>/30<sup>th</sup>, 2020)

TimothyGu:
To test another time zone, use the `timeZone` option (that defaults to your current time zone):
```js
new Intl.DateTimeFormat('en-US', { timeZone: 'Asia/Kolkata', timeZoneName: 'short' }).formatToParts(new Date('2020-01-01')).slice(-1)[0].value
new Intl.DateTimeFormat('en-US', { timeZone: 'Asia/Kolkata', timeZoneName: 'long' }).formatToParts(new Date('2020-01-01')).slice(-1)[0].value
```

PR #60, opened 2020-07-09
Timezone-proof website

Timthetic:
Fix: #57, #65


Timthetic:
I have not changed the banner yet. To do that affectively, we'd need a start time as well. I'm also partially unsure about if we should change the banner based on time zone and *not* display the *time* the event starts nearby.

jamieliu386:
What do you mean by needing a start time?

I think it would suffice to make the date on the left reflect the days the event will span for a person locally. (i.e. Feb 23rd/24th if it spans two days, or simply Feb 23rd or Feb 24th if spans a single day)

I don't think we need to change anything about the countdown.

src/components/HomePage/Banner.js
```diff
@@ -90,11 +117,19 @@ function renderInfo(classes) {
...
+					`It looks like HOTH crosses between dates in your timezone! (${tz})` :
+					''}
+				placement='top'
+				arrow='true'>
+					<Typography variant='h5' className={classes.text} component='h3'>
+						<time dateTime={hothStart.toISOString()}>
+							{month} {startDay}<sup>{ordinal(startDay)}</sup>{endDayElement}, 2021
+						</time>
+					</Typography>
+				</Tooltip>
```
Timthetic, line 131: I thought I'd throw in the tooltip and see what you all thought. It provides a bit more explanation to the user.

src/components/HomePage/Banner.js
```diff
@@ -90,11 +117,19 @@ function renderInfo(classes) {
 					width: 150,
 					margin: 20
 				}} />
-			<Typography variant='h5' className={classes.text} component='h3'>
-				<time dateTime={hothStart.toISOString()}>
-					Feb 23<sup>rd</sup>, 2021
-				</time>
-			</Typography>
+			<Box className={classes.horizontalBox}>
```
jamieliu386, line 120: I think we can get rid of the `horizontalBox` class and specify the styles here.
```suggestion
			<Box display='flex' alignItems='center' color='white'>
```

src/components/HomePage/Banner.js
```diff
@@ -29,6 +46,11 @@ const useStyles = makeStyles(theme => ({
 			padding: '12px 2px'
 		}
 	},
+	horizontalBox: {
+		display: 'flex',
+		alignItems: 'center',
+		color: 'white'
+	},
```
jamieliu386, line 53: See my other comment
```suggestion
```

src/components/HomePage/Banner.js
jamieliu386: I like the tooltip!

src/components/HomePage/Banner.js
```diff
@@ -74,6 +78,11 @@ const useStyles = makeStyles(theme => ({
 }));
 
 function renderInfo(classes) {
+	const [{ value: month },, { value: tz }] = timeFormatter.formatToParts(hothStart);
+	const startDay = hothStart.getDate();
+	const endDay = hothEnd.getDate();
+	const eventCrossesDate = startDay !== endDay;
+	const endDayElement = eventCrossesDate ? <>–{endDay}</> : <></>;
```
TimothyGu, line 85:
```suggestion
	const endDayString = eventCrossesDate ? `–${endDay}` : '';
```

src/components/HomePage/Banner.js
```diff
@@ -7,12 +7,16 @@ import Box from '@material-ui/core/Box';
...
 
 import SvgImg from '../SvgImg';
 import hothLogo from '../../images/hoth7-logo.svg';
 import hothBanner from '../../images/hoth-banner.svg';
 
-const hothStart = new Date('2021-02-23T09:00:00-07:00');
+const hothStart = new Date('2021-02-23T07:00:00-07:00');
+const hothEnd = new Date('2021-02-23T21:00:00-07:00');
+const timeFormatter = new Intl.DateTimeFormat('en-US',
+	{ timeZoneName: 'short', month: 'short' });
```
TimothyGu, line 19:
```suggestion
const timeFormatter = new Intl.DateTimeFormat('en-US', { timeZoneName: 'short', month: 'short' });
```

src/components/HomePage/Banner.js
```diff
@@ -74,6 +78,11 @@ const useStyles = makeStyles(theme => ({
 }));
 
 function renderInfo(classes) {
+	const [{ value: month },, { value: tz }] = timeFormatter.formatToParts(hothStart);
```
TimothyGu, line 81:
```js
const monthFormatter = Intl.DateTimeFormat('en-US', { month: 'short' }).format(new Date());
monthFormatter.format(hothStart);
```

src/components/SchedulePage/TimeSlot.js
```diff
@@ -18,18 +21,22 @@ const useStyles = makeStyles({
...
 		<>
 			<Divider />
 			<Box display='flex'>
 				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage} />
 				<div>
 					<Typography component='h2'>
-						<time dateTime={time.toISOString()}>{timeFormatter.format(time)}</time>
+						<time dateTime={time.toISOString()}>
+							{timeFormatter.format(time).toString()}
+							<Box component='span' paddingLeft='1em'>
```
Timthetic, line 36: I'm still unsure about the kind of space we want to add here (how much).

src/components/SchedulePage/TimeSlot.js
```diff
@@ -18,18 +21,23 @@ const useStyles = makeStyles({
...
 			<Box display='flex'>
 				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage} />
 				<div>
 					<Typography component='h2'>
-						<time dateTime={time.toISOString()}>{timeFormatter.format(time)}</time>
+						<time dateTime={time.toISOString()}>
+							{timeFormatter.format(time).toString()}
+							<Box color='gray' component='span'>
+								{' '}
+								{currentTimeZoneShort.value}
```
TimothyGu, line 38: Oops, I didn't realize the `.value` was needed. My intention was so that `timezone_names` would just export strings. You should move the `.value` there.

src/components/SchedulePage/TimeSlot.js
```diff
@@ -18,18 +21,23 @@ const useStyles = makeStyles({
...
 	return (
 		<>
 			<Divider />
 			<Box display='flex'>
 				<ScheduleRoundedIcon fontSize={'small'} className={classes.clockImage} />
 				<div>
 					<Typography component='h2'>
-						<time dateTime={time.toISOString()}>{timeFormatter.format(time)}</time>
+						<time dateTime={time.toISOString()}>
+							{timeFormatter.format(time).toString()}
```
TimothyGu, line 35: Can you check if the `.toString()` is needed? I feel it might not be.

src/components/SchedulePage/TimeSlot.js
```diff
@@ -18,18 +21,23 @@ const useStyles = makeStyles({
 	}
 });
 
-const timeFormatter = new Intl.DateTimeFormat('en-US', { timeStyle: 'short', hour: 'numeric', minute: '2-digit' });
-
 function Timeslot({ time, events }) {
 	const classes = useStyles();
+	timeFormatter.formatToParts(time);
```
TimothyGu, line 26: Unused.

src/components/HomePage/Banner.js
```diff
@@ -74,6 +78,12 @@ const useStyles = makeStyles(theme => ({
 }));
 
 function renderInfo(classes) {
+	const tz = currentTimeZoneShort;
+	const month = monthFormatter.format(hothStart);
+	const startDay = hothStart.getDate();
+	const endDay = hothEnd.getDate();
+	const eventCrossesDate = startDay !== endDay;
```
TimothyGu, line 85: Maybe add a comment here that `startDay` and `endDay` are in the user's timezone?

src/components/HomePage/Banner.js
```diff
@@ -7,12 +7,16 @@ import Box from '@material-ui/core/Box';
...
 import Countdown from 'react-countdown';
+import Tooltip from '@material-ui/core/Tooltip';
+import { currentTimeZoneShort } from '../SchedulePage/timezone_names.js';
 
 import SvgImg from '../SvgImg';
 import hothLogo from '../../images/hoth7-logo.svg';
 import hothBanner from '../../images/hoth-banner.svg';
 
-const hothStart = new Date('2021-02-23T09:00:00-07:00');
+const hothStart = new Date('2021-02-23T07:00:00-07:00');
```
TimothyGu, line 17: HOTH starts at 7am? I thought it was 9am.

src/components/SchedulePage/TimeSlot.js
Timthetic: We don't, format() returns a string. Updated.

src/utils/timezone_names.js
```diff
@@ -0,0 +1,12 @@
+function getTimeZoneWithFormat(format) {
+	try {
+		const formatter = new Intl.DateTimeFormat('en-US', { timeZoneName: format });
+		return formatter.formatToParts(new Date()).find(part => {
+			return part.type === 'timeZoneName';
+		}) || '';
```
TimothyGu, line 6:
```suggestion
		return formatter.formatToParts(new Date()).find(part => {
			return part.type === 'timeZoneName';
		}).value || '';
```

and remove the `.value` below. Right now, you would be trying to get `.value` from a string sometimes, which is not quite what we want.

Issue #39, opened 2020-06-20
We need a favicon


TimothyGu:
Right now, we are still using Gatsby's default favicon. We should change it so that our own logo is used.

It's up to us which logo we should use: Hack's, or design a HOTH logo.

https://github.com/uclaacm/hoth.uclaacm.com/blob/cb22f273981aae54cbf495448619ab2ab3f0d501/gatsby-config.js#L27

Timthetic:
Just for experimentation, I took screenshots with the hack logo and the logo from hoth7. I like the Hack logo a lot because it is simple, so it looks nice even though it is small. If we ask design to create one, it should be simple unlike the Hoth7 logo (which I'm sure was not designed with the intention of being used as a favicon). One idea is to have it be inspired off of the Hack logo, but look slightly different or more "intense".

<img width="1280" alt="Website-with-hack-logo" src="https://user-images.githubusercontent.com/33593205/85893476-cc1b8b00-b7a7-11ea-9716-1ce16a7c39fd.png">
<img width="1279" alt="Website-with-hoth-logo" src="https://user-images.githubusercontent.com/33593205/85893485-cf167b80-b7a7-11ea-8c39-90e02d0aeb8c.png">

jamieliu386:
I think we should use the Hack logo for now because the current splashthat site uses it, and like you said, it's much cleaner than the hoth7 logo.

We can always design or get a new favicon in the future if we decide we want something different.

TimothyGu:
Hack logo sounds good. We definitely should not use the HOTH logo, though I was hoping someone could come up with a simplified design for the HOTH logo.

PR #61, opened 2020-07-10
Change favicon to hack logo

milesswu:
Hoth website favicon updated from default gatsby icon to use our hack logo in
accordance with #39


gatsby-config.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
 				background_color: '#fff',
 				theme_color: '#c960ff',
 				display: 'minimal-ui',
-				icon: 'src/images/gatsby-icon.png' // This path is relative to the root of the site.
```
TimothyGu, line 32: You should remove this file.

milesswu: Oh! Good catch, thank you. Resolved in the next commit!

PR #63, opened 2020-07-19
Test PR: add code

TimothyGu:


.github/workflows/lint.yml
```diff
@@ -1,16 +0,0 @@
-name: Lint
```
TimothyGu, line 1: To reviewers: ignore this change! It's done in order to prevent ESLint errors from showing up on the PR page.

src/components/GalleryPage/Gallery.js
```diff
@@ -0,0 +1,50 @@
...
+						}
+					}
+				}
+			}
+		}
+	`);
+
+	const image= []
+
+	for (let i =0; i<=data.carouselImages.nodes;i++) {
```
TimothyGu, line 38: 1. The index should be compared against the length of the array, rather than the array itself.

2. Because JavaScript is 0-indexed, we should use `<` to compare the index and the length rather than `<=`.

```suggestion
	for (let i = 0; i < data.carouselImages.nodes.length; i++) {
```

src/components/GalleryPage/Gallery.js
```diff
@@ -0,0 +1,50 @@
...
+						fluid {
+							...GatsbyImageSharpFluid
+						}
+					}
+				}
+			}
+		}
+	`);
+
+	const image= []
```
TimothyGu, line 36: This is an array variable; let's make the name plural to indicate that, for readability.

```suggestion
	const images = [];
```

src/components/GalleryPage/Gallery.js
```diff
@@ -0,0 +1,50 @@
...
+
+	for (let i =0; i<=data.carouselImages.nodes;i++) {
+		let node = data.carouselImages.nodes[i]
+		image.push(<Img fluid={{ ...node.childImageSharp.fluid, aspectRatio: 2 }} key={node.id} />);
+	}
```
TimothyGu, line 41: You can actually use `data.carouselImages.nodes.map` function, which can simplify the code a bit.

src/components/GalleryPage/Winners.js
```diff
@@ -0,0 +1,60 @@
...
+			width: '100%',
+			padding: '10px 0px'
+		}
+	}
+}));
+
+function Winners() {
+	const classes = useStyles()
+
+	const winner=winners.map(w=><p>{w.category}: {w.description}</p>)
```
TimothyGu, line 50: Rename it to `winnerEntries`, which indicates that this is an array, and makes it more specific and different from the `winners` array declared above.

Issue #81, opened 2020-08-30
Consider more consistency in font size on the homepage


TimothyGu:
From @GalenWong 

Upcoming Workshops fonts are much smaller. Consider using the regular body font.

TimothyGu:
@milesswu to investigate possible options

milesswu:
increase font size of workshop title to balance larger workshop description font size

PR #87, opened 2020-08-30
Adjustments to upcoming workshops styles

milesswu:
Fixes: #81 

Description text made more consistent, using theme's 'body2' styles. 
Title text made larger to balance the larger description.


jamieliu386 (approved):
I agree with @TimothyGu that the font size change for the title isn't 100% necessary, but I think it's nice for it to be consistent with with the FAQ title, so I'm also fine with keeping the change.

Issue #84, opened 2020-08-30
New Schedule Page Design


milesswu:
Redesign for schedule page. See [Figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=15%3A9)

However, want to keep groupings (keep sticky time slot and add dividers between groupings).

Issue #77, opened 2020-08-30
Add duration to the workshop schedule


TimothyGu:
From feedback from @khxia and @GalenWong

TimothyGu:
* If event is remote, then we can replace the location information / room with the duration
* If event is not remote, then we could place the duration information as part of the description

PR #89, opened 2020-08-31
Schedule page redesign

milesswu:
Fixes: #84, #77 

Schedule page format redesigned as per [Figma](https://www.figma.com/file/sC4tsO37gn2qgn1ol1yfZq/HOTH-Website?node-id=650%3A6)

Uses the same Event.js component for both the HomePage and SchedulePage. Subtitle items turned into an array of strings to handle placing max two pieces of information on every line.

Also added duration to events.


src/components/HomePage/Event.js
```diff
@@ -20,32 +19,51 @@ const useStyles = makeStyles(theme => ({
...
 		marginTop: '1ex',
 		'& > :nth-child(1)': {
-			marginRight: '1.5em'
+			marginRight: '1em'
+		},
+		[theme.breakpoints.up('sm')]: {
+			'& > :nth-child(1)': {
+				marginRight: '1.5em'
+			}
 		}
```
TimothyGu, line 28: Consider

```suggestion
		'& > :nth-child(1)': {
			marginRight: '1em'
			[theme.breakpoints.up('sm')]: {
				marginRight: '1.5em'
			}
		}
```

also consider using `:nth-child(odd)`, considering we want margin-right on subtitles 1, 3, 5, … (though it doesn't currently matter since we never have 4 subtitles)

src/components/HomePage/Event.js
```diff
@@ -5,7 +5,6 @@ import Box from '@material-ui/core/Box';
 import Grid from '@material-ui/core/Grid';
 import Typography from '@material-ui/core/Typography';
 
-const timeFormatter = new Intl.DateTimeFormat('en-US', { hour: 'numeric', minute: '2-digit', timeZoneName: 'short' });
```
TimothyGu, line 8: remove empty line:

```suggestion
```

src/components/HomePage/Event.js
```diff
@@ -20,32 +19,51 @@ const useStyles = makeStyles(theme => ({
...
+	const chunkSize = 2;
+
+	// split subtitles array into chunks of 2 (only two items should be shown on the same line of any subtitle)
+	for (let i = 0; i < subtitles.length; i += chunkSize) {
+		const subtitleChunk = subtitles.slice(i, i + chunkSize);
+		const renderChunk = subtitleChunk.map((subtitle, index) => {
+			return (
+				<Typography variant='body2' key={`${name}-subtitle-${index}`}>
+					{subtitle}
+				</Typography>);
```
TimothyGu, line 51:
```suggestion
				</Typography>
			);
```

src/components/SchedulePage/Schedule.js
```diff
@@ -5,28 +5,40 @@ import Typography from '@material-ui/core/Typography';
...
+		<Typography align='left' component='h1' variant='h4' className={classes.title}>Schedule</Typography>
 		<List>
-			{workshopSchedule.map(timeslot => {
+			{workshopSchedule.map((timeslot, index) => {
 				return (
 					<React.Fragment key={timeslot.startTime}>
-						<Divider component='li' />
 						<li>
 							<StickyTimeSlot events={timeslot.events} time={timeslot.startTime} />
+							{index === workshopSchedule.length - 1 ? null : <Divider className={classes.divider} />}
```
TimothyGu, line 41: Why change this to be different from what we had? I.e., `<Divider component='li' />` instead of putting Divider in another li.

src/data/WorkshopSchedule.js
```diff
@@ -13,6 +13,7 @@ const workshopSchedule = [
 		events: [
 			{
 				name: 'ACM-W × ACM Hack | Hackathon 101 + Team Formation',
+				duration: '30 min',
```
TimothyGu, line 16: OPTIONAL

This works, though you can store the duration as an integer here, and then use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) to render the number as text.

If you decide to go down this path, make sure to change the property key to `durationMin`, so that the unit of the number is clear.

```js
const formatter = new Intl.NumberFormat('en-US',  {
	style: 'unit',
	unit: 'minute'
});
formatter.format(10); // returns '10 min'
```

src/components/HomePage/Event.js
```diff
@@ -57,9 +75,8 @@ function Event({ startTime, name, location, description }) {
 }
 
 Event.propTypes = {
-	startTime: PropTypes.instanceOf(Date).isRequired,
 	name: PropTypes.string.isRequired,
-	location: PropTypes.string.isRequired,
+	subtitles: PropTypes.array.isRequired,
```
TimothyGu, line 79: nit:
```suggestion
	subtitles: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
```

Ideally we would make this component work even if subtitles is not supplied, but since we don't need that capability right now it's fine to make `subtitles` a required prop.

src/components/SchedulePage/StickyTimeSlot.js
```diff
@@ -27,6 +27,8 @@ const useStyles = makeStyles(theme => {
 			}
 		},
 		eventItem: {
+			paddingTop: theme.spacing(3),
+			paddingBottom: theme.spacing(8),
```
TimothyGu, line 31: Sorry, this is deep into bikeshedding territory, but 

```suggestion
			marginTop: theme.spacing(2),
			marginBottom: theme.spacing(6),
```

looks a bit better to me. padding → margin, since margins collapse (the margin-bottom of the previous item and the margin-top of the next item overlap). The spacing numbers are just bikeshedding.

TimothyGu (approved):

jamieliu386 (approved):

src/data/WorkshopSchedule.js
milesswu: Pin for a later time, not necessary right now

Issue #78, opened 2020-08-30
Look into increasing image quality


TimothyGu:
@GalenWong noticed that the quality of the images are quite lackluster. We should look into improving them, by increasing JPEG compression quality parameter, using a better format like WebP, and/or increasing resolution.

PR #96, opened 2020-10-01
Improve default image quality

jamieliu386:
Fixes: #78 

Changes `gatsby-config.js` to increase the default quality value of images from 50 to 96. Additionally modifies the GrqphQL queries to use `_withWebP` as specified by the [gatsby-image documentation](https://www.gatsbyjs.com/plugins/gatsby-image/) in order to take advantage of the more efficient [WebP image format](https://en.wikipedia.org/wiki/WebP).

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -18,7 +18,12 @@ module.exports = {
 			}
 		},
 		'gatsby-transformer-sharp',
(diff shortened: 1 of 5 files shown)


TimothyGu (approved):
:shipit:

Issue #97, opened 2021-01-05
Change instances of "Hack on the Hill" to "Hack off the Hill"


Timthetic:
- [ ] Bonus points: Use [sed](https://www.gnu.org/software/sed/manual/sed.html) to do this automatically.

TimothyGu:
sshhhhhh this is supposed to be secret or something right

PR #113, opened 2021-01-10
Switch on to off in Hack on the hill

curesh:
Fixes: #97
I used `find . -type f -exec sed -i '' -e "s/Hack\ on\ the\ Hill/Hack\ off\ the\ Hill/g" {} \;` from the ./src folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 <img align=left width=100 src=src/images/hack-logo.svg alt="Hack logo">
 
-# Hack on the Hill (HOTH) Website
+# Hack off the Hill (HOTH) Website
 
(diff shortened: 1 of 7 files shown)


TimothyGu:
A few notes on the command

- Don't use `find` for Git repos, since they include `.git` as well as generated/ignored files (as you have probably found out). Instead use `git ls-files | xargs sed ...`
- The `-i ''` thing only works on macOS (and other BSD-derived systems). On Linux, it's `-i` (without `''`)

TimothyGu (changes requested):
A couple more places to fix
- README.md
- package.json
- gatsby-config.js

TimothyGu:
Also, can you add `Fixes: #97` to the PR description?

Issue #118, opened 2021-01-16
Sticky header on schedule page shows up above the main hamburger menu


TimothyGu:
![image](https://user-images.githubusercontent.com/1538624/104823539-d9b66200-5818-11eb-977b-b772f6003662.png)

PR #121, opened 2021-01-16
Increase z-index of mobile menu bar

TimothyGu:
The sticky header of the schedule page also has z-index: 1, causing conflicts with the menu bar.

Fixes: #118


jamieliu386 (approved):

jodymlin (approved):
You're a 10, just like the new z-index 👍

PR #129, opened 2021-01-23
Updated prize page

### LINKED ISSUE #128: Prizes page outdated
Prizes still need to be determined, hide info from the site for now.

jakobreinwald:
Fixes #128 : Labels old prizes as old


src/components/PrizePage/Prizes.js
```diff
@@ -157,7 +157,7 @@ export default function Prizes() {
 											{prize.prize}
 										</Typography>
 									</Box>
-									<Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} />
+									{/* <Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} /> */}
```
TimothyGu, line 160: This works, but what may be a bit better is: On lines 16–52, remove the `img` properties to denote that we don't have images for those prizes. Then, here, we can then check to see the prize has an image, and only try to display it if that's the case.

```js
{prize.img && <Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} />}
```

The advantage of this approach is that we wouldn't have to figure out why the images are not showing up in the future.

TimothyGu (approved):
Looks good; one suggestion.

src/components/PrizePage/Prizes.js
jakobreinwald: I just tried this and if I delete the img properties of the prizes and add this line, the header images are also removed. So the circles above each prize will also be left empty. Is that okay?

TimothyGu: Ah never mind, I see it. LGTM.

TimothyGu (approved):

PR #132, opened 2021-01-23
Hid Schedule page

### LINKED ISSUE #131: Schedule page not up to date
Hide schedule info from the page for now.

jakobreinwald:
Fixes #131 : I just commented out the code that displayed the schedule because I figured we would just need to put it right back once we have the updated schedule, but I can delete it if you want.

### DIFF
--- a/src/components/SchedulePage/Schedule.js
+++ b/src/components/SchedulePage/Schedule.js
@@ -3,17 +3,17 @@ import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
 import Container from '@material-ui/core/Container';
-import List from '@material-ui/core/List';
-import Divider from '@material-ui/core/Divider';
+// import List from '@material-ui/core/List';
+// import Divider from '@material-ui/core/Divider';
 
-import StickyTimeSlot from '../SchedulePage/StickyTimeSlot';
-import workshopSchedule from '../../data/WorkshopSchedule';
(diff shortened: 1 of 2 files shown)


src/components/SchedulePage/Schedule.js
```diff
@@ -3,17 +3,17 @@ import React from 'react';
...
 import Container from '@material-ui/core/Container';
-import List from '@material-ui/core/List';
-import Divider from '@material-ui/core/Divider';
+// import List from '@material-ui/core/List';
+// import Divider from '@material-ui/core/Divider';
 
-import StickyTimeSlot from '../SchedulePage/StickyTimeSlot';
-import workshopSchedule from '../../data/WorkshopSchedule';
+// import StickyTimeSlot from '../SchedulePage/StickyTimeSlot';
+// import workshopSchedule from '../../data/WorkshopSchedule';
```
TimothyGu, line 11: As an alternative to commenting out code (which is hard to maintain), you could do the following. Add a line called

```suggestion

const scheduleAvailable = false;
```

Then, later change the `paddingBottom` line to

```
		paddingBottom: theme.spacing(scheduleAvailable ? 4 : 11)
```

and then change `<List ...> ... </List>` to

```
{scheduleAvailable && <List ...>
  ...
</List>}
```

TimothyGu (commented):
Thanks! One suggestion

TimothyGu:
Oh, another comment. We should hide the workshop listing on the home page as well.

TimothyGu:
I pushed a commit that addresses both comments. Once again, nice work on this!

Issue #95, opened 2020-10-01
Banner redesign


jamieliu386:
The banner needs to be redesigned.

Adding to the backlog for now since we'll probably need new assets from design when the time comes.

jamieliu386:
We got assets! No longer blocked :p

jamieliu386:
https://www.figma.com/file/X7ZuxzntCJwG6Ae8kGztcm/HOTH-8?node-id=0%3A1

PR #135, opened 2021-01-30
HOTH8 Banner Redesign

christinatong01:
Fixes: #95


src/components/HomePage/Banner.js
```diff
@@ -81,41 +97,42 @@ const useStyles = makeStyles(theme => ({
...
 					`It looks like HOTH crosses between dates in your timezone! (${tz})` :
 					''}
 				placement='top'
 				arrow={true}>
-					<Typography variant='h5' className={classes.text} component='h3'>
-						<time dateTime={hothStart.toISOString()}>
-							{month} {startDay}{endDayString}, 2021
-						</time>
+					<Typography variant='h5' className={classes.text} style={{ paddingLeft: 57, paddingTop: 40 }} component='h3'>
+							February 5 - 7, 2021 | Virtual
```
jamieliu386, line 127: We used variables here because the start/end date may differ for people in other time zones.

Also, we should keep the [`time` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) here since that provides more information for screen readers.

src/components/HomePage/Banner.js
```diff
@@ -26,9 +26,9 @@ const useStyles = makeStyles(theme => ({
 		height: 'auto',
 		[theme.breakpoints.up('md')]: {
 			backgroundImage: `url(${hothBanner})`,
-			backgroundSize: '545px 329px',
+			backgroundSize: '1200px 500px',
```
jamieliu386, line 29: It seems like this is why the banner is unresponsive for small screens. We should probably maintain the split between the left portion with the text/info and the right section with the graphic and countdown

src/components/MenuBar/ButtonBar.js
```diff
@@ -69,7 +69,7 @@ function ButtonBar({ isMobile }) {
 				<PoppinLink to={link.to} key={`nav-${index}`}>
 					{link.name}
 				</PoppinLink>)}
-			<BorderLink href='https://forms.gle/7uokDycPQfU9B5oj8' target='_blank'>
+			<BorderLink href={'https://forms.gle/7uokDycPQfU9B5oj8'} target='_blank'>
```
jamieliu386, line 72: This is a remnant of the previous branch you based this branch off of, we should remove the change from here.

jamieliu386 (changes requested):
Left some comments for you to review, but don't worry about fixing the issues as I'll take care of them today.

src/components/HomePage/Banner.js
```diff
@@ -81,41 +97,42 @@ const useStyles = makeStyles(theme => ({
...
-			justify='center'
-			alignItems='center'
+			justify='left'
+			alignItems='left'
 		>
-			<Typography variant='h3' className={classes.text} component='h2' style={{ fontSize: '3em' }}>
-				Hack off the Hill 8
+			<img src = {hothTitle}
+			style={{ paddingLeft: 50, paddingBottom: 20, width: '80%' }}
+			/>
```
TimothyGu: We don't need to use an image for the text title. According to the Figma, the title is just the standard Poppins font. Additionally, you can export the gradient on the "8" in CSS format by clicking on the "Copy" button here:

image

That puts the following CSS string onto the system clipboard:
```css
background: linear-gradient(199.69deg, #FF5F96 35.64%, #BD01FF 95.25%);
```

That's not entirely enough to make the gradient work for text, but [this post](https://cssgradient.io/blog/css-gradient-text/) documents the other two properties you need to set:

```css
-webkit-background-clip: text;
-webkit-text-fill-color: transparent;
```

Check out Jamie's modifications for how it all fits together.

src/components/HomePage/Banner.js
TimothyGu: In general, the SVG we export should only contain the parts that we cannot reproduce ourselves in HTML. That means we don't need to export the (large) background rectangle either.

TimothyGu (approved):

PR #136, opened 2021-02-03
Working schedule page

### LINKED ISSUE #100: Update Schedule
🛑 **Blocked** on Hoth workshop schedule completion.

Schedule should be updated to include opening/closing events, office hour schedule, etc.

curesh:
Fixes #100

Based on schedule provided here: https://docs.google.com/spreadsheets/d/1hHP0p-YDF8RUTpSWt3HGAOwJn8NkqP2rnApadVNuKeg/edit#gid=0

### DIFF
--- a/src/components/HomePage/Event.js
+++ b/src/components/HomePage/Event.js
@@ -51,9 +51,7 @@ function Event({ name, subtitles, description }) {
 					</Typography>
 				);
 			}
-			return (
-				<></>
-			);
+			return null;
 		});
 		renderSubtitle.push(
 			<Box className={classes.subtitle}>
--- a/src/components/HomePage/Event.js
+++ b/src/components/HomePage/Event.js
@@ -44,10 +44,15 @@ function Event({ name, subtitles, description }) {
 	for (let i = 0; i < subtitles.length; i += chunkSize) {
 		const subtitleChunk = subtitles.slice(i, i + chunkSize);
 		const renderChunk = subtitleChunk.map((subtitle, index) => {
+			if (subtitle !== '') {
+				return (
+					<Typography variant='body2' key={`${name}-subtitle-${index}`}>
+						{subtitle}
+					</Typography>
+				);
+			}
 			return (
-				<Typography variant='body2' key={`${name}-subtitle-${index}`}>
-					{subtitle}
-				</Typography>
+				<></>
 			);
 		});
 		renderSubtitle.push(
--- a/src/components/HomePage/UpcomingEvents.js
+++ b/src/components/HomePage/UpcomingEvents.js
@@ -84,7 +84,7 @@ function UpcomingEvents() {
 	return <Box component="section" paddingY={{ xs: 8, md: 10 }} bgcolor='background.grey'>
 		<Container maxWidth='md' className={classes.eventsContainer}>
 			<Typography component='h2' variant='h4'
-				className={classes.title}>Upcoming Workshops</Typography>
+				className={classes.title}>Upcoming Events</Typography>
 			{renderEvents}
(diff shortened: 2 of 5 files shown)


src/components/SchedulePage/Schedule.js
```diff
@@ -35,7 +35,7 @@ const SchedulePage = () => {
 	const classes = useStyles();
 	return <Container maxWidth='md'>
 		<Typography align='left' component='h1' variant='h4' className={classes.title}>
-			Schedule Coming Soon!
+			Events Schedule
```
jamieliu386, line 38: Can you change line "Upcoming Workshops" on line 87 of this file to say "Upcoming Events"?

src/components/HomePage/UpcomingEvents.js
```diff
@@ -10,9 +10,10 @@ import workshopSchedule from '../../data/WorkshopSchedule';
 import Event from './Event';
 
 // When changing this, make sure to update src/components/SchedulePage/Schedule.js as well.
-const scheduleAvailable = false;
+const scheduleAvailable = true;
```
jamieliu386, line 13: Can you change line "Upcoming Workshops" on line 87 of this file to say "Upcoming Events"?

src/components/HomePage/Event.js
```diff
@@ -44,10 +44,15 @@ function Event({ name, subtitles, description }) {
...
+						{subtitle}
+					</Typography>
+				);
+			}
 			return (
-				<Typography variant='body2' key={`${name}-subtitle-${index}`}>
-					{subtitle}
-				</Typography>
+				<></>
 			);
```
jamieliu386, line 56: Replace this with `return null;`

src/components/SchedulePage/Schedule.js
jamieliu386, line 38:
```suggestion
			Schedule
```

src/data/WorkshopSchedule.js
```diff
@@ -1,75 +1,186 @@
-import React from 'react';
-
-const hothDate = '2021-02-23'; // replace with real date later
+const hothDate = '2021-01-'; // replace with real date later
 const utcOffset = '-08:00'; // PST
 // time is a string of the form "HH:mm:ss"
 function getDateObj(time) {
-	return new Date(`${hothDate}T${time}${utcOffset}`);
+	return new Date(`${hothDate}${time}${utcOffset}`);
```
jamieliu386, line 5: If you can, just modify the function to also take in the day (i.e. `05`, `06`, etc.), and rewrite the function body to mimic the original. The new function should take a `day` and a `time` parameter. (Likewise, modify the places where you called the function to pass in two arguments).

While the solution you came up with works and is very clever, it creates a leak in the abstraction, as the user has to know how this function is implemented to use it.

src/data/WorkshopSchedule.js
```diff
@@ -1,75 +1,186 @@
-import React from 'react';
-
-const hothDate = '2021-02-23'; // replace with real date later
+const hothDate = '2021-01-'; // replace with real date later
```
jamieliu386, line 1: This is the wrong month. Also, it'd be nice to rename this variable to represent more accurately the data it contains.
```suggestion
const hothMonth = '2021-02-';
```

jamieliu386 (changes requested):
Nice work overall, but a couple things need fixes!

jamieliu386 (approved):
Nice work, I just caught one more thing: In the Upcoming Events on the homepage, could you change the button text to say something like "More Events" instead of "More Workshops"? Other than that, I think this is good to go :)

image

PR #139, opened 2021-02-05
Jakob/newprizes

### LINKED ISSUE #103: Update prizes page with actual prizes
Once the prizes have been decided (also waiting on #76), update the prizes page to reflect that.

jakobreinwald:
Fixes #103 

- Added images to all the prizes and changed from last year's to this year's


src/components/PrizePage/PrizeHeader.js
```diff
@@ -30,7 +30,7 @@ export default function PrizeHeader() {
 				Prize Announcements Coming Soon!
```
TimothyGu, line 30:
```suggestion
				Prize Tracks
```

src/components/PrizePage/PrizeHeader.js
```diff
@@ -30,7 +30,7 @@ export default function PrizeHeader() {
...
 			</Typography>
 			<Typography variant='h3' component='h2' className={classes.lastYear}>
-				Last Year’s Prizes:
+				This Year’s Prizes:
 			</Typography>
```
TimothyGu, line 34: We should now delete this and the `lastYear` class.

src/components/PrizePage/Prizes.js
```diff
@@ -157,7 +157,7 @@ export default function Prizes() {
 											{prize.prize}
 										</Typography>
 									</Box>
-									{/* <Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} /> */}
+									{<Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} /> }
```
TimothyGu, line 160:
```suggestion
									<Img fixed={prizeImageMap.get(prize.img)} style={{ height: '128px' }} />
```

TimothyGu (changes requested):

jamieliu386 (dismissed):
In addition to what @TimothyGu already mentioned:

At a certain screen width, I get the below view. To fix it, we should have a transition between the three cards side by side to the single card view where there's two cards per row.

image

Also, this card looks a bit off because the text is too wide, so we should add some side padding to the text to make it wrap.

image

TimothyGu:
I also see you created a `package-lock.json` file. That means you used npm to develop this repository, which is not what we use (we use Yarn instead).

TimothyGu (approved):

PR #141, opened 2021-02-05
Fix workshop page container width

jodymlin:
The current workshop page is wrapped in a `Container` of size `lg` right now. Noticed because it causes awkward horizontal overflow on my laptop.

The rest of the website uses `md` for `Container` so I just changed it. 

Not sure if this fix is in another upcoming PR, but I didn't see any issue made for it, so I made this one just in case. Feel free to reject if another PR was going to fix this.

### DIFF
--- a/src/components/GalleryPage/Winners.js
+++ b/src/components/GalleryPage/Winners.js
@@ -83,10 +83,12 @@ function Winners() {
 				<hgroup>
 					<Typography variant='h5' component='h2'>{item.title}</Typography>
 					<Typography variant='subtitle1' component='h3'
-						style={{ fontWeight: theme.typography.fontWeightRegular,
+						style={{
 							textTransform: 'uppercase',
 							fontSize: '1em',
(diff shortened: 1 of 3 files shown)


jamieliu386 (approved):

TimothyGu:
This doesn't actually fix the bug of page overflow:

image

I also like having the videos be wider when possible…

TimothyGu (changes requested):
See above

jodymlin:
@TimothyGu

I actually prefer it smaller because its too big for my laptop screen with the container maxWidth set to large: 
image

With the medium container its much more readable. 

There's a way to fix the overflow (I'm pretty sure it has to do with the way Material-UI does padding for their grid items). I most likely won't get to it today since I'm still helping Asha wrap up some last minute HOTH prep, but I can try get a fix up at least sometime tomorrow if it can't be done today.

TimothyGu (approved):
Just fixed the overflow, and also some other assorted problems. LGTM.

Issue #20, opened 2020-05-30
Reminder to update packages once in a while


TimothyGu:

jamieliu386:
Steps:
1. run `yarn`
2. run `yarn outdated` to see which packages to update. example:
   ```
   Package                    Current Wanted  Latest  Package Type    URL                                                                                      
   @material-ui/core          4.10.0  4.10.2  4.10.2  dependencies    https://material-ui.com/
   eslint                     7.1.0   7.2.0   7.2.0   devDependencies https://eslint.org
   gatsby                     2.22.13 2.23.3  2.23.3  dependencies    https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby#readme
   ```
3. run `yarn add <dependencies...>` where `<dependencies...>` refer to a space-separated list of `dependencies` shown as outdated in the step above. example:
  `$ yarn add @material-ui/core gatsby`
4. run `yarn add --dev <devDependencies...>` where `<devDependencies...>` refer to a space-separated list of `devDependencies` shown as outdated in step 2. example:
  `$ yarn add --dev eslint`
5. run `yarn upgrade`
6. run `yarn develop` and make sure everything still works

If you run into weird issues after updating packages, try the steps outlined in: https://github.com/uclaacm/hoth.uclaacm.com#troubleshooting

PR #152, opened 2021-12-21
upgrade node version to v16

khxia:
Upgraded the required node version to `v16.x` in order to align with the Hack website.

#20


jodymlin (changes requested):
Update README but other than that LGTM

jodymlin (approved):

Issue #165, opened 2022-01-08
Add Announcements page


jodymlin:
Create a new page that contains any announcements for/during HOTH.

Rather than having announcements created by directly editing the a file in the repo, making a PR, and merging, we are thinking of using something similar to the [serverless comment system](https://www.gatsbyjs.com/blog/2018-04-10-how-to-handle-comments-in-gatsby-blogs/) Gatsby recommends.

khxia:
Use a Github Action. 
Triggered on the issue_comment event: 
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#issue_comment 

Payload example:
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment 

make sure to filter to allow only team/organization

Run a script that calls the Github Issue API - list comment issues: https://docs.github.com/en/rest/reference/issues#list-issue-comments 
to get all comments and then inject into file

jodymlin:
Let's break the PR's for this into smaller tasks so each person can take on a small piece

- [ ] Create static Announcements page
- [ ] Create an announcement component and list on page
- [ ] Create script to call Github Issue API to get all comments on an issue
- [ ] Create Github Action workflow that triggers on the issue_comment event

PR #171, opened 2022-01-14
Nathan/new announcements page

nathanzzhang:
- Created static Announcements page with text "Check back here for announcements!"
- Added Announcements tab to menu bar

#165

### DIFF
--- a/src/components/AnnouncementsPage/Announcements.js
+++ b/src/components/AnnouncementsPage/Announcements.js
@@ -2,19 +2,20 @@ import React from 'react';
 
 import Typography from '@material-ui/core/Typography';
 import useTheme from '@material-ui/core/styles/useTheme';
+import Container from '@material-ui/core/Container';
 
 export default function AnnouncementsPage() {
 	const theme = useTheme();
 
 	return (
-		<>
+		<Container>
 			<Typography variant='h4' component='h1'
 				style={{ fontWeight: theme.typography.fontWeightBold,
-					paddingTop: theme.spacing(37), // initial value 4
-					paddingBottom: theme.spacing(37), // initial value 4
+					paddingTop: theme.spacing(4),
+					paddingBottom: theme.spacing(4), 
 					textAlign: 'center' }}>
 				Check back here for announcements!
 			</Typography>
-		</>
+		</Container>
 	);
 }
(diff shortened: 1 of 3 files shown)


src/components/AnnouncementsPage/Announcements.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default function AnnouncementsPage() {
+	const theme = useTheme();
+
+	return (
+		<>
+			<Typography variant='h4' component='h1'
+				style={{ fontWeight: theme.typography.fontWeightBold,
+					paddingTop: theme.spacing(37), // initial value 4
+					paddingBottom: theme.spacing(37), // initial value 4
```
jodymlin, line 14: I can see that you used this to vertically center the text on the screen and also make text fill up the entire screen (so the footer doesn't look awkward in the middle of the screen). Screen sizes differ, however, and this results in the page sometimes taking up too much height and sometimes still taking up too little height. 

I just made new fix that makes all pages take up a minimum height of the size of the screen (regardless of how much content is on the page) so you don't need to have this large padding anymore. Feel free to just use the original `theme.spacing(4)` padding. It's not a big deal if the text is not vertically centered.

jodymlin (changes requested):
Some minor changes requested but otherwise great job!

The fix I mentioned about the screen height is in pr #174 so don't worry about having to manually fill the page space.

src/components/AnnouncementsPage/Announcements.js
```diff
@@ -0,0 +1,21 @@
...
+
+import Typography from '@material-ui/core/Typography';
+import useTheme from '@material-ui/core/styles/useTheme';
+import Container from '@material-ui/core/Container';
+
+export default function AnnouncementsPage() {
+	const theme = useTheme();
+
+	return (
+		<Container>
```
jodymlin, line 11:
```suggestion
		<Container maxWidth='md'>
```

All the other pages use `md` width containers, so let's make this `md` as well for consistency. This prop and its values can be found [here](https://mui.com/api/container/) on the MUI docs for `Container`

jodymlin (changes requested):
One last minor change to make then looks good to me and I'll approve it 👍

src/components/AnnouncementsPage/Announcements.js
nathanzzhang: ah i see, thanks!

jodymlin (approved):

AnakinTrotter (approved):
LGTM!

Issue #167, opened 2022-01-13
Remove Facebook Link


jodymlin:
Remove this section from homepage

![image](https://user-images.githubusercontent.com/44387428/149276383-c3666f4a-8f35-4c97-ba98-e70041e83c93.png)

Krish-bhardwaj:
Hi would like to contribute to your project can u assign me this issue

PR #172, opened 2022-01-15
Removed Facebook link

AnakinTrotter:
Removed Facebook link component from the home page. 

#167


jodymlin:
Looked good to me but make sure to get 2 approvals before merging.

PR #174, opened 2022-01-16
Header/Footer takes up full page

jodymlin:
Before, if there wasn't enough content on the page, the footer would high up on the page, rather than on the bottom.

This fix makes the page always take up _at least_ the height of a screen, regardless how much content is on the page.


khxia (approved):
LGTM, although seeing another `makeStyle` used hurts my heart because we need to get rid of it eventually.

PR #177, opened 2022-01-18
Updated gallery page with last year's winners

jakobreinwald:
Updated gallery page with winners from last year and added pictures of those projects into images/gallery-winners


src/components/GalleryPage/Winners.js
```diff
@@ -11,40 +11,59 @@ import useMediaQuery from '@material-ui/core/useMediaQuery';
...
-		title: 'Bruin Dining',
-		category: 'Best Design, Best React Native',
-		description: `A more beautiful and insightful way to dine on the Hill. It's
-			no fun going to a dining hall and being surprised or disappointed by the
-			food.`,
-		link: 'https://devpost.com/software/bruin-dining-table-16',
-		id: 'bruindining'
+		title: 'Mammal Mayhem',
+		category: 'Best Game',
+		description: `Mammal Mayhem is a cutesy game designed to provide casual relief to its players.`,
```
khxia, line 56: The description for this is a bit short, which means there's a small alignment inconsistency with "Clothes with Care". If there was no particular order when making this, could you move the projects around so that 'Mammal Mayhem' is right next to 'EmoSign' (which only has three lines)? Of course make sure to preserve the ordering such that the two 'Best Web Hacks' are still next to each other.

khxia (commented):
This looks fine to me, just a small nitpick about the alignment

src/components/GalleryPage/Winners.js
jakobreinwald: Sounds good, I put EmoSign and Mammal Mayhem as the bottom pair and then Reassure(You Do You) as the sole one on the bottom. The alignment between EmoSign and Mammal Mayhem isn't perfect still but it's a lot closer than before.

maggieelli (approved):
Looks great to me!

khxia (approved):
LGTM!

Issue #158, opened 2022-01-08
Add HOTH 9 application link


jodymlin:
Link the HOTH 9 application link

PR #179, opened 2022-01-21
Added Application link and made it so that it will become available on Monday morning 

tmag1:
I updated the link for applications and made it so that the apply button will be available from the soft launch to the application deadline. #158

### DIFF
--- a/src/components/HomePage/Banner.js
+++ b/src/components/HomePage/Banner.js
@@ -143,7 +143,7 @@ function renderInfo(classes) {
 				</Tooltip>
 			</Box>
 			{Date.now() < applyDeadline.getTime() &&
-				<Button disabled className={classes.apply} href={'https://forms.gle/7uokDycPQfU9B5oj8'} target='_blank'>
+				<Button disabled className={classes.apply} href={'https://forms.gle/YVuiEyQZo5VT82E68'} target='_blank'>
 					Apply Now
 				</Button>
 			}
--- a/src/components/HomePage/Banner.js
+++ b/src/components/HomePage/Banner.js
@@ -143,8 +143,11 @@ function renderInfo(classes) {
 					</Typography>
 				</Tooltip>
 			</Box>
-			{Date.now() > applicationOpen.getTime() && Date.now() < applyDeadline.getTime() &&
-				<Button className={classes.apply} href={'https://forms.gle/YVuiEyQZo5VT82E68'} target='_blank'>
+			{
+				<Button
+					disabled={Date.now() < applicationOpen.getTime() || Date.now() > applyDeadline.getTime()}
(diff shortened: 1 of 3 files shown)


src/components/MenuBar/ButtonBar.js
```diff
@@ -77,13 +81,14 @@ function ButtonBar({ isMobile }) {
 				<PoppinLink to={link.to} key={`nav-${index}`}>
 					{link.name}
 				</PoppinLink>)}
+			{Date.now() > applicationOpen.getTime() &&
 			<BorderLink
-				disabled
```
jodymlin, line 81: Rather than making the button on the Navigation Bar disappear, let's just disable it.

The original disabled code was this
```
disabled={Date.now() > applyDeadline.getTime()}
```

So modify that as necessary with the application open day

jodymlin (changes requested):

jodymlin:
Also be sure to add a description to the pull request for good practice and link the corresponding issue on the side so it will close the issue when the PR is merged.

src/components/MenuBar/ButtonBar.js
```diff
@@ -4,6 +4,10 @@ import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
 import { Link } from 'gatsby';
 
+import {
```
khxia, line 7:
```suggestion
import { applicationOpen } from '../constants.js';
```
```

**Review by khxia — CHANGES_REQUESTED** [review]

**Review by khxia — APPROVED** [review]
LGTM! I think the button on the banner looks weird when it's disabled. I think maybe in the future, we could have a different color scheme based on when it's disabled or not, but for now this will do.

edit: It seems that we want the one on the banner to disappear and the nav bar one to be disabled. Since you will be busy right now, I will do it

File: src/components/HomePage/Banner.js [hunk 3]
```diff
@@ -142,8 +143,11 @@ function renderInfo(classes) {
 					</Typography>
 				</Tooltip>
 			</Box>
-			{Date.now() < applyDeadline.getTime() &&
-				<Button disabled className={classes.apply} href={'https://forms.gle/7uokDycPQfU9B5oj8'} target='_blank'>
+			{
+				<Button
+					disabled={Date.now() < applicationOpen.getTime() || Date.now() > applyDeadline.getTime()}
```
khxia, line 148: Instead of disabling this one, can you make it disappear instead?
We want the navbar one to be disabled, but not the banner one.

khxia (changes requested):

khxia (approved):
LGTM

jodymlin (approved):
LGTM 👍 Good work!

Issue #157, opened 2022-01-08
Update main graphics


jodymlin:
Update the main HOTH logo and front banner with the new graphics provided by [ACM Design](https://trello.com/b/4RWnKdxv/acm-design-requests).

PR #180, opened 2022-01-21
Nathan/update graphics

nathanzzhang:
Updated home page banner with Hoth9 graphic 

#157 

image


src/components/HomePage/Banner.js
```diff
@@ -28,7 +28,7 @@ const useStyles = makeStyles(theme => ({
 	container: {
 		backgroundColor: theme.palette.primary.dark,
 		height: 'auto',
-		padding: 25,
+		padding: 0,
```
jodymlin, line 31:
```suggestion
```
Padding by default on divs (which this style is applied to) is already 0, so we can just remove this line.

src/components/HomePage/Banner.js
```diff
@@ -151,6 +153,23 @@ function renderInfo(classes) {
 	);
 }
 
+const Image = () => {
+	const data = useStaticQuery(graphql`
+    query {
+      placeholderImage: file(relativePath: { eq: "powellBackground.png" }) {
+        childImageSharp {
+          fluid(maxWidth: 300) {
```
jodymlin, line 161:
```suggestion
          fluid(quality: 100) {
```
When we specify the maxWidth, the PNG quality shrinks a lot and looks really fuzzy when you zoom in. Let's remove the maximum width parameter and also specify the image quality to be 100 so we get that the highest PNG quality possible.

Unfortunately this is just one of the downsides of PNG compared to SVG. Although the colors have retained the original design better, you'll notice that when you zoom in it's fuzzy. If you use an SVG however (for example the banner on the hack website), you'll notice that you do not get fuzzy edges when you zoom in.

src/images/hoth9-logo.svg
```diff
@@ -0,0 +1,159 @@
+<svg width="512" height="512" viewBox="0 0 512 512" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jodymlin, line 1: You can remove this file. I'll be including it in my PR so you don't need to add it.

src/components/HomePage/Banner.js
```diff
@@ -68,8 +68,10 @@ const useStyles = makeStyles(theme => ({
 		justifyContent: 'center',
 		alignItems: 'baseline',
 		alignSelf: 'flex-end',
+		padding: '25px 0px 0px 0px',
```
jodymlin, line 71: For spacing, we like to use the `theme.spacing()` function so we get consistent units and if we ever need to scale up or down the spacing we'll be able to do it across all the files in `layout.js` by overriding the default style.

The default spacing in material UI is 8px, so let's turn this into the closest factor of 8 with `theme.spacing()`, which would be 24px. (3 * 8 = 24px)
```suggestion
		padding: theme.spacing(3, 0, 0, 0),
```

src/components/HomePage/Banner.js
```diff
@@ -68,8 +68,10 @@ const useStyles = makeStyles(theme => ({
...
 		[theme.breakpoints.down('sm')]: {
-			alignSelf: 'flex-start'
+			alignSelf: 'flex-start',
+			padding: '15px'
```
jodymlin, line 74: Similar fix for padding. Use `theme.spacing()` and set to spacing to the closest factor of 8 (which is 16px). If you need a refresher on how to convert `16px` padding to `theme.spacing()`, check out the [spacing MUI Docs](https://mui.com/customization/spacing/).

src/components/HomePage/Banner.js
```diff
@@ -151,6 +153,23 @@ function renderInfo(classes) {
 	);
 }
 
+const Image = () => {
```
jodymlin, line 156: Rather than storing the component into an `Image` variable, let's just move the static query inside the `Banner` function component. You can place it right next to all the other `const` variables inside the `Banner` component.

It will look something like
```js
function Banner() {
  const data = useStaticQuery(...);
  
  return ... // stuff that makes up the Banner component here
}
```

See my other suggestion below on how to inline the `` component.

src/components/HomePage/Banner.js
```diff
@@ -151,6 +153,23 @@ function renderInfo(classes) {
...
+        childImageSharp {
+          fluid(maxWidth: 300) {
+            ...GatsbyImageSharpFluid
+          }
+        }
+      }
+    }
+  `);
+
+	return <Img fluid={data.placeholderImage.childImageSharp.fluid} />;
```
jodymlin, line 169: Rather than returning `` here we'll just inline this in the `Banner` component. See the suggestion below

src/components/HomePage/Banner.js
```diff
@@ -220,13 +239,13 @@ function Banner() {
...
 						<Grid item sm={12} md={6}>
 							<Box display='flex' flexDirection='column'>
-								<SvgImg src={hothBanner} className={classes.logo} width={936} height={581} />
 								<NoSsr>
 									<Countdown
 										date={hothStart}
 										renderer={countdownRenderer}
 									/>
 								</NoSsr>
+								<Box>{Image}</Box>
```
jodymlin, line 248:
```suggestion
								<Box>
								  <Img fluid={data.placeholderImage.childImageSharp.fluid} />
								</Box>
```

Inline the `` component here.

jodymlin (changes requested):

jodymlin (approved):
LGTM 😄 great job!

maggieelli (commented):
Looks good!

PR #184, opened 2022-01-25
Changed remaining "Hack off" to "Hack on"

jodymlin:
Removed all remaining instances of "Hack Off" in the site metadata and changed it to "Hack On"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 <img align=left width=100 src=src/images/hack-logo.svg alt="Hack logo">
 
-# Hack off the Hill (HOTH) Website
+# Hack on the Hill (HOTH) Website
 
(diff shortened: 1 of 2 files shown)


khxia (approved):
yeet

TimothyGu:
One more:

https://github.com/uclaacm/hoth.uclaacm.com/blob/d60e4b929f0db77459f10dd69a2119f639be673d/package.json#L4

khxia:
Oops, will fix now

jodymlin:
Oh my god I swear if we go virtual and we have to change it back to "off" I'm going to throw hands 😭 

But also ty for the catch I must've filtered out the file in VSCode on accident

PR #185, opened 2022-01-25
Add Github Actions for Announcements Page

khxia:
- Added a Github Actions workflows that triggers when someone comments on issue number 181
- Also created a dummy Python script that will eventually be modified to fetch and inject all comments from issue number 181.

#165

### DIFF
--- a/.github/workflows/announcements.yml
+++ b/.github/workflows/announcements.yml
@@ -27,9 +27,9 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install -r .github/python_requirements.txt
-      - name: Run updateAnnouncements script
+      - name: Run update_announcements script
       - run: |
-        python .github/actions/python_requirements
+          python ./.github/actions/update_announcements.py
       - uses: stefanzweifel/git-auto-commit-action@v4
         name: Commit and push
         with:
--- a/.github/workflows/announcements.yml
+++ b/.github/workflows/announcements.yml
@@ -25,8 +25,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
       - name: Install Python dependencies
-        env:
-          ISSUE_NUMBER: ${{ github.event.issue.number }}
         run: |
           python -m pip install -r .github/python_requirements.txt
       - name: Run update_announcements script
--- a/.github/workflows/announcements.yml
+++ b/.github/workflows/announcements.yml
@@ -8,11 +8,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 16.x
-        uses: actions/setup-node@v1
-        with:
-          node-version: '16.x'
-      - run: yarn install --frozen-lockfile
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
@@ -28,7 +23,7 @@ jobs:
         run: |
           python -m pip install -r .github/python_requirements.txt
       - name: Run update_announcements script
-      - run: |
+        run: |
           python ./.github/actions/update_announcements.py
       - uses: stefanzweifel/git-auto-commit-action@v4
         name: Commit and push
--- a/.github/workflows/announcements.yml
+++ b/.github/workflows/announcements.yml
@@ -4,7 +4,12 @@ jobs:
   issue_commented:
     # This job only runs for issue comments
     name: Update Announcements
-    if: ${{ !github.event.issue.pull_request }}
+    if: |
+      !github.event.issue.pull_request &&
+      github.event.issue.number == 181 &&
+      (github.event.issue.comment.author_association == "OWNER" ||
+      github.event.issue.comment.author_association == "MEMBER" ||
+      github.event.issue.comment.author_association == "CONTRIBUTOR")
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/announcements.yml
```diff
@@ -0,0 +1,35 @@
+name: Fetch announcements from issue
+on: issue_comment
+jobs:
+  issue_commented:
+    # This job only runs for issue comments
+    name: Update Announcements
+    if: ${{ (!github.event.issue.pull_request) && github.event.issue.number == 181 }}
```
jodymlin, line 7: Because the variables inside the `${{ }}` are substituted with their values and then executed they are subject to an injection attack. See the [Security Hardening guides](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).

Although I don't think in our case anyone will be able to inject anything into `github.event.issue.number` it might just be good practice to move it into an [environment variable](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).

.github/workflows/announcements.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: |
+          python -m pip install -r .github/python_requirements.txt
+      - name: Run update_announcements script
+        run: |
+          python ./.github/actions/update_announcements.py
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        name: Commit and push
+        with:
+          commit_message: "Updated Announcements"
+          file_pattern: src/*.js src/*.json
```
jodymlin, line 32: Again, not entirely sure if there are security risks, but since we are force pushing, to follow the principle of least privilege, let's change this file pattern to be limited to only the announcements data file.

khxia: Yes, for now, not sure what the format will be like, I guess I will make it a json for now

.github/workflows/announcements.yml
khxia: Do you mean move `181` as an environment variable, or do we pass in the `github.event.issue.number` as an environment variable into the python script

jodymlin: Is it also possible to also limit this to trigger if someone we trust has posted an issue? Otherwise random users can comment on the issue and trigger this workflow, which will eventually trigger a build and eat up our build quota.

I found an [example payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-example-when-someone-edits-an-issue) of the `github.event` payload when someone updates an issue. Looks like we'll get some metadata about the user who made the change. Maybe the `site_admin` property will help?

jodymlin: > Do you mean move `181` as an environment variable, or do we pass in the `github.event.issue.number` as an environment variable into the python script

`github.event.issue.number`. These are the variables that will be substituted and can thus be substituted by some bash script. In this case I don't think anyone can change the `number` into some string that gets executed by bash, but just in case.

khxia: So if we use a whitelist, the python script will not make any changes to any files. The commit action will then not commit anything to the repo. So even though the workflow will trigger, we will not be building anything. 

The `site_admin` property though is a good double-check

khxia: Ok I don't really like this new method:
Because the script will still run when any change is made on any issue -> although nothing will happen in unrelated issues, it still feels bad because we can do the check much earlier

.github/workflows/announcements.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/cache@v2
+        name: configure pip caching
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/python_requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Install Python dependencies
+        env:
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
```
jodymlin, line 29: We settled on not using this in an environment variable right? We can take this out since it's not being used anymore

jodymlin (changes requested):

jodymlin (approved):

mattxwang:
Just wanted to say, this is a very cool way to do this. Any reason why y'all chose this over something else (ex markdown file, JSON, etc.)?

khxia:
Main reason we didn't go with a markdown file/JSON was that then we would need to create a new branch, make PR, then merge. Using github actions, we can actually automate all of that - we were trying to not use a CMS and still be able to get a 'ui' that we can add comments to, and then it will just magically appear on our website, even though it's static. 

This is also the simplest and most maintainable way we came up with doing things after wracking our brains for a while. There definitely could be better ways of doing things, but we're small brain. 😅

mattxwang:
Super cool stuff! Something that is maybe a long-term goal for ACM's dev team (assuming priorities don't change) is building a similar infra for the main ACM Website, and I agree that any file-based mechanism with PRs can be a bit cumbersome (especially to those w/ no technical experience). Maybe this'll work out bigger picture!

I do wonder - could you add a layer of safeguard security by making someone write the message in an issue body with a certain label/tag, and then updating the file with a bot on merge? Creative Labs does similar things with their [RFCs repo](https://github.com/UCLA-Creative-Labs/cl-rfcs).

Learn something new every day :)

khxia:
Will definitely look into it. Thanks!

PR #186, opened 2022-01-25
Changed the final 'hack on the hill' to 'hack off the hill'

khxia:
We missed an 'off the hil' in #184 . This change is addressing that


jodymlin (approved):
i cri

PR #192, opened 2022-01-28
Created Announcement component, list, and JSON file

maggieelli:
Linked to #165


src/components/AnnouncementsPage/Announcement.js
```diff
@@ -0,0 +1,45 @@
...
+function Announcement({ subject, timestamp, body }) {
+	const classes = useStyles();
+
+	return (
+		<Container>
+			<Typography variant='h4' component='h2' className={classes.subject}>
+				{subject}
+			</Typography>
+			<Typography variant='h6' component='h3' className={classes.timestamp}>
+                Posted on: {timestamp}
```
khxia, line 30: Let's do the timestamp stuff here.
So normally the timestamp data will be passed in like some sort of string: `"2021-01-21T08:13:11Z"`

In JavaScript, we can use the built-in `Date` object.
```js
let date = new Date("2021-01-21T08:13:11Z");
date.getDate();
...
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date 

That way we should be able to format the string into the format that we want.

khxia (changes requested):

src/components/AnnouncementsPage/Announcements.js
```diff
@@ -14,8 +16,20 @@ export default function AnnouncementsPage() {
 					paddingTop: theme.spacing(4),
```
khxia, line 16: To make things more 'future-proof', let's use `makeStyles` here too.

khxia (commented):
Looks good to me, you can change `announcements.json` to an empty array now.

src/components/AnnouncementsPage/Announcements.js
```diff
@@ -1,21 +1,43 @@
...
+						key={announcement.id}
+						subject={announcement.subject}
+						timestamp={announcement.timestamp}
+						body={announcement.body}
+					/>
+					{announcement.id === data.length ?
+						null :
+						<Divider
+						/>}
+				</>)};
```
khxia, line 40: Get rid of the semicolon, I think it's rendering on the screen

khxia (changes requested):

khxia (approved):

PR #202, opened 2022-02-06
Fix Github Actions Error

khxia:
Found a mistake that I made in #185 . This should fix it. Also added a dummy Python script for better testing.


jodymlin (approved):
I'm going to trust you on this one

PR #203, opened 2022-02-07
Jakob/issueScript

jakobreinwald:
Created a python script to get all of the issues/announcements from issue 181 (where we will be listing our announcements) and insert them into a json file to be put into our announcements page.


khxia (changes requested):

updateAnnouncements.py
```diff
@@ -0,0 +1,42 @@
...
+    if(element['user']['login'] in user_whitelist):
+        body = element['body']
+        subject, partition, comment = body.partition('(Subject) ')
+        comment = {'id': index, 'subject': subject, 'body': comment, 'timestamp': element['created_at']}
+        valid_comments.append(comment)
+
+# Insert all valid comment json objects into a file
+filename = 'src/data/json/announcements.json'
+json_string = json.dumps(valid_comments, indent=4)
+with open(filename, 'w') as f:
```
khxia, line 39: The ingenuity here is that even if we re-write the exact same json into the file, it won't trigger a commit and build because we technically didn't change anything.

updateAnnouncements.py
```diff
@@ -0,0 +1,42 @@
...
+print("Request completed with status code: ", announcements.status_code)
+announcements_json = announcements.json()
+
+valid_comments = []
+
+# Keep track of every comment from valid usernames
+for index, element in enumerate(announcements_json):
+    if(element['user']['login'] in user_whitelist):
+        body = element['body']
+        subject, partition, comment = body.partition('(Subject) ')
```
khxia, line 32: Have you also tested the behaviour of what adding new lines `'\n'` does? You might also need to deal with CRLF and LF (`'\r\n'` vs `'\n'`) differences in Mac and Windows. 

Lmk if you're confused on how to deal with it.

updateAnnouncements.py
```diff
@@ -0,0 +1,52 @@
...
+
+# Convert announcements from Github API to json for parsing
+try:
+    announcements_json = announcements.json()
+except:
+    print("Error obtaining json from Github API return. ")
+valid_comments = []
+
+# Keep track of every comment from valid usernames
+try:
```
khxia, line 36: Let's remove the try-catch here, and let the program fail on its own when we encounter an error.

updateAnnouncements.py
```diff
@@ -0,0 +1,52 @@
...
+            subject, partition, comment = body.partition('(Subject) ')
+            comment = {'id': index, 'subject': subject, 'body': comment, 'timestamp': element['created_at']}
+            valid_comments.append(comment)
+except:
+    print("Error parsing issue comments. ")
+
+# Insert all valid comment json objects into a file in reverse order
+filename = 'src/data/announcements.json'
+json_string = json.dumps(valid_comments[::-1], indent=4)
+with open(filename, 'w') as f:
```
khxia, line 49: Let's also add a try-catch here, since writing to a file could fail.

Refer to this post: https://stackoverflow.com/questions/713794/catching-an-exception-while-using-a-python-with-statement

khxia (changes requested):
Looks good to me! Just a few more requests on try catches, after I did more research.

khxia:
@jakobreinwald You'll also need to merge again from main because I merged something into main that should cause merge conflicts

khxia:
From what we talked about on the messenger chat, could add do `body.trim()` in `Announcement.js`?

khxia:
LGTM! Final changes before we merge:
1. Delete all the comments from the issue and make sure annountements.json is empty `[]`
2. Rename `updateAnnouncements.py` to `update_announcements.py` and get rid of the current `update_announcements.py`

maggieelli (approved):
Looks good!

khxia (approved):
LGTM

Issue #161, opened 2022-01-08
Paginate Gallery page


jodymlin:
Going to the Gallery page should, by default, go to the most recent winner's page. However, there should be links to go to the past HOTH winners.

A few things about each new page
* each new page corresponds to a previous HOTH
* if there are no images of the winners specifically for a HOTH, we will not display a photo carousel at the top

jodymlin:
- [ ] Create YAML file structure that represents gallery page data
- [ ] Create page query in `gatsby-node.js`
- [ ] Update GalleryPage component to use data passed in via `pageContext`
  -  remove Gallery page from static page folder and build via `createPages` API

PR #215, opened 2022-02-18
Anakin/paginate gallery page

AnakinTrotter:
Past winners are now viewable at /gallery/hoth-i where i is a number from 1 to 8.
There is also a dropdown menu to access past HOTHs.

#161 #218


khxia:
Are you planning on having the carousel be the same across different years?

jakobreinwald (approved):
Looks good to me!

gatsby-node.js
```diff
@@ -1,7 +1,60 @@
...
+			winnersArray.push(winnerMap);
+		}
+		hothWinners.set(hoth.parent.name, winnersArray);
+		devposts.set(hoth.parent.name, hoth.parent.childYaml.devpost);
+	}
+	return { hothWinners, devposts };
+}
+
+exports.createPages = async ({ actions: { createPage }, graphql }) => {
+	const numPages = 8;
```
jodymlin, line 21: Let's not hardcode this value

src/components/GalleryPage/GalleryPageTemplate.js
```diff
@@ -0,0 +1,27 @@
...
+			<Gallery />
+			<Winners winners={ winnerInfo } devpost={ devpostInfo } />
+		</Layout>
+	);
+}
+
+GalleryPageTemplate.propTypes = {
+	pageContext: PropTypes.shape({
+		winnerInfo: PropTypes.array.isRequired,
+		devpostInfo: PropTypes.string
```
jodymlin, line 23: Why is this not required?

src/components/MenuBar/ButtonBar.js
```diff
@@ -69,7 +69,7 @@ function ButtonBar({ isMobile }) {
 		},
 		{
 			name: 'Gallery',
-			to: '/gallery'
+			to: '/gallery/hoth-8'
```
jodymlin, line 72: I'm wondering if there's a better way than hardcoding this. I don't have ideas at the moment and if we don't think of any this is okay. Just make a comment somewhere that whenever we add new HOTH data we have to update this as well.

src/data/winners/hoth-2.yml
```diff
@@ -0,0 +1,34 @@
...
+    - title: Mood for Thought
+      category: Best Design Hack
+      description: "user sends mood/personalized note around the world; another user \"grabs\" note and sends back personalized song rec"
+      link: "https://devpost.com/software/mood-for-thought"
+      image: anon
+    - title: Cook Up a Recipe
+      category: Best Web Hack
+      description: Easy Link to Recipes
+      link: "https://devpost.com/software/cook-up-a-recipe"
+      image: cookuparecipe
```
jodymlin, line 34: nit: add a new line at the end of the file

jodymlin (changes requested):
Really good work! This is a great PR :)

Is it possible to change the button text from "See all Projects" to "See all HOTH # Projects"? And then for each page instead of "Past Winners" as the title, we can say "HOTH # Winners"?

src/components/GalleryPage/GalleryPageTemplate.js
```diff
@@ -0,0 +1,27 @@
...
+import Gallery from '../../components/GalleryPage/Gallery';
+import Winners from '../../components/GalleryPage/Winners';
+
+function GalleryPageTemplate({ pageContext }) {
+	const { winnerInfo, devpostInfo } = pageContext;
+	return (
+		<Layout>
+			<SEO title='Gallery' />
+			<Gallery />
+			<Winners winners={ winnerInfo } devpost={ devpostInfo } />
```
khxia, line 15: Better name is `devpostLink`?

khxia (changes requested):
Minor changes LGTM

khxia:
Also, boop on the `yarn` `npm` thing as a reminder

src/components/GalleryPage/GalleryMenu.js
```diff
@@ -0,0 +1,58 @@
...
+				TransitionComponent={Fade}
+			>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-8'>HOTH 8</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-7'>HOTH 7</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-6'>HOTH 6</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-5'>HOTH 5</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-4'>HOTH 4</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-3'>HOTH 3</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-2'>HOTH 2</Link></MenuItem>
+				<MenuItem onClick={handleClose}><Link to='/gallery/hoth-1'>HOTH 1</Link></MenuItem>
```
jodymlin, line 54: Let's loop this instead of listing each out individually

jodymlin (changes requested):
Very minor change requested, but otherwise looks great :)

gatsby-node.js
```diff
@@ -1,7 +1,62 @@
...
+            }
+          }
+	  }
+    `);
+	const { hothWinners, devposts } = getHothWinners(hothData.data.allYaml.nodes);
+	for (let i = 1; i <= hothWinners.size; i++) {
+		const hothName = `hoth-${i}`;
+		const winnerInfo = hothWinners.get(hothName);
+		const devpostLink = devposts.get(hothName);
+		const hothNum = i;
```
TimothyGu, line 52: Rather than constructing `hothName` from an index, I'd just loop through all the keys of `hothWinners`, like so:

```js
for (const hothName of hothWinners.keys() {
	const winnerInfo = hothWinners.get(hothName);
	const devpostLink = devposts.get(hothName);
	// ...
```

If you decide to adopt this style, a natural question would then be, what should `hothNum` be set to? For that, I actually suggest that you _remove_ `hothNum`, and change the GalleryPageTemplate and Winners components to accept a `hothName` string instead of `hothNum`. You can then populate `hothName` using the `name` field inside the YAML files. Not only does this proposal allow for more elegant iteration, but (more importantly) it also makes the code more flexible. As an example, it's not unthinkable for future Hack to skip HOTH 9 (like Microsoft Windows and iPhones), or name HOTH 10 as "HOTH X" (like iPhones and macOS). Your current code would be too rigid to handle that case.

khxia (changes requested):

src/components/GalleryPage/GalleryMenu.js
```diff
@@ -0,0 +1,62 @@
...
+	const open = Boolean(anchorEl);
+	const handleClick = event => {
+		setAnchorEl(event.currentTarget);
+	};
+	const handleClose = () => {
+		setAnchorEl(null);
+	};
+	const menuLinks = [];
+	for (let i = hothCount; i >= 1; i--) {
+		menuLinks.push(<MenuItem onClick={handleClose}><Link to={`/gallery/hoth-${i}`}>HOTH {i}</Link></MenuItem>);
```
TimothyGu, line 21: This is missing a `key={…}` prop.

TimothyGu (commented):
Hello, I [used to](https://github.com/uclaacm/hoth.uclaacm.com/commits?author=TimothyGu) work on this code base and wanted to give a few suggestions for your consideration.

Aside from the inline code comments, I also suggest that you try to organize the `gallery-winners/` by each HOTH, so you have photos in each of `gallery-winners/hoth-4/`… As more HOTHs happen, it would make people's lives much easier. (Imagine digging through hundreds of photos to find something, all in the same directory…)

gatsby-node.js
AnakinTrotter: Thank you, Timothy. I will get right on this.

khxia:
I think this comment

> Aside from the inline code comments, I also suggest that you try to organize the gallery-winners/ by each HOTH, so you have photos in each of gallery-winners/hoth-4/… As more HOTHs happen, it would make people's lives much easier. (Imagine digging through hundreds of photos to find something, all in the same directory…)

by tim hasn't been addressed yet

gatsby-node.js
```diff
@@ -1,7 +1,73 @@
...
+		const winnerInfo = hothWinners.get(key);
+		const devpostLink = devposts.get(key);
+		const galleryLink = allGalleryLinks.get(key);
+		const hothNames = Array.from(allHothNames.values());
+		const galleryLinks = Array.from(allGalleryLinks.values());
+		// This is an absurd number of props and there is probably a much
+		// better way to do this that somebody can figure out at a later time
+		// On the bright side it is now easier to change names and links
+		createPage({
+			path: `${galleryLink}`,
```
khxia, line 62: Potentially, we could also get rid of the `allGalleryLinks` map . 
In the path, we would pass in `gallery/${key}`, and we would just have the `galleryLinks = Array.from(hothNames.keys())`

Not sure if this semantically makes more sense though

gatsby-node.js
```diff
@@ -1,7 +1,73 @@
...
+          }
+	  }
+    `);
+	const { hothWinners, devposts, allHothNames, allGalleryLinks } = getHothWinners(hothData.data.allYaml.nodes);
+	for (const key of hothWinners.keys()) {
+		const hothName = allHothNames.get(key);
+		const winnerInfo = hothWinners.get(key);
+		const devpostLink = devposts.get(key);
+		const galleryLink = allGalleryLinks.get(key);
+		const hothNames = Array.from(allHothNames.values());
```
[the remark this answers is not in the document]
AnakinTrotter: I am going to move it out of the loop for now.

khxia (approved):

khxia (approved):
LGTM

PR #224, opened 2022-02-18
Fixed MenuBar collapse threshold

jakobreinwald:
Changed the threshold for isMobile so that the MenuBar collapses before the buttons get crunched and a second row is created.


nathanzzhang (approved):
nice!

jodymlin (approved):

PR #230, opened 2022-02-18
Fix sticky time slot offset

### LINKED ISSUE #226: Sticky headers on Schedule page
Currently, the headings for each of the time segments of the schedule page are sticky to their sections, but there's an awkward gap between the top of the page and the actual sticky header.

the issue with sticky header offset

Two possible solutions:
- Make headers non-sticky
- Fix header to stick to the top or add background to that blank space (unsure of this implementation)

I don't think the sticky header design is crucial so this is probably a design choice you can make.

khxia:
Fixes #226


jodymlin (approved):

khxia:
On my screen, it looks like all the

PR #233, opened 2022-02-18
Change Announcement partition

### LINKED ISSUE #232: Update partition for Announcement comments
Currently we enforce that a user _must_ follow this format
```
Subject (Subject) Body
```
which enforces a single space after the word `(Subject)` partition. This however is not intuitive to users who might not be familiar with python code, and in particular, how we use the `partition()` function. 

Rather than force the user to understand our implementation, we can make this more flexible by changing the partition from `"(Subject) "` (with a space) to `"(Subject)"` (w/out a space) and call `.strip()` on the tuple outputs. User comment input will now be more flexible and less prone to accidental formatting issues that cause unexpected Announcement output.

jodymlin:
Change partition from `"(Subject) "` to `"(Subject)"`

Fixes #232

### DIFF
--- a/update_announcements.py
+++ b/update_announcements.py
@@ -36,9 +36,9 @@ valid_comments = []
 for index, element in enumerate(announcements_json):
     if element['user']['login'] in user_whitelist:
         body = element['body']
-				# If there is no partition, subject will contain the entire string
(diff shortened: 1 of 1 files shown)


khxia (approved):
yeet

PR #239, opened 2022-02-22
Fix accessibility of Apply Button

### LINKED ISSUE #229: Accessibility of Apply Button
image

Should have white text and no underline when hovering over it

jodymlin:
Make text white and remove underline of Apply Button on HomePage

Fixes #229

### DIFF
--- a/src/components/MenuBar/ButtonBar.js
+++ b/src/components/MenuBar/ButtonBar.js
@@ -15,11 +15,11 @@ const useStyles = makeStyles(theme => {
 		},
 		borderBtn: {
 			margin: 10,
+			color: '#FFFFFF',
 			transition: theme.transitions.create(['color', 'background'], {
 				duration: theme.transitions.duration.complex
 			}),
 			'&:hover': {
-				color: '#FFFFFF',
 				background: '#DB99FD'
 			},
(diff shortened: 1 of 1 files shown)


khxia (commented):
This looks good to me. But why is the button disabled on mobile mode? And then when you switch back from mobile mode back to desktop mode, it's still disabled. Is this the same weird bug that we encountered before?

nathanzzhang (changes requested):
Same bug as Alex pointed out, but other than that, color looks great!

jodymlin:
> This looks good to me. But why is the button disabled on mobile mode? And then when you switch back from mobile mode back to desktop mode, it's still disabled. Is this the same weird bug that we encountered before?

@khxia 
I didn't run into this issue when I initially developed locally, but just now when I built it again the button was disabled on both mobile and desktop versions. Still, I don't think the styling has anything to do with it (it's definitely something wacky with the `disabled` prop date comparison. 

```
disabled={Date.now() < applicationOpen.getTime() || Date.now() > applyDeadline.getTime()}
```

I say for now, if it happens again when we merge this branch we can open an issue to fix it.

khxia (approved):

nathanzzhang (approved):

PR #271, opened 2022-04-10
Upgrade gatsby version from 2.24.65 to 2.24.80

khxia:
We had an error where our newly added package: `gatsby-transformer-yaml` was not compatible with our old version of gatsby.
```txt
Plugin gatsby-transformer-yaml is not compatible with your gatsby version 2.24.65 - It requires gatsby version 2.24.80.
```
Since, it was a minor minor version upgrade, I decided to quickly stop the error from appearing by upgrading our gatsby version to 2.24.80.
In the future, we should definitely take the time to upgrade our major packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@material-ui/styles": "^4.10.0",
     "@svgr/webpack": "^5.4.0",
     "classnames": "^2.2.6",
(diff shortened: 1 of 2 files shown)


katelynsyu (approved):
and poof the error is gone :)

PR #314, opened 2023-01-21
improved styling for dropdown of past winners

datowq:
might add lines to separate the drop-down items better, but PRing for advice on where to put the styling

### DIFF
--- a/src/components/GalleryPage/GalleryMenu.js
+++ b/src/components/GalleryPage/GalleryMenu.js
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 import useTheme from '@material-ui/core/styles/useTheme';
 import { makeStyles } from '@material-ui/core/styles';
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles(theme => ({
 	linkStyle: {
 		color: theme.palette.primary.dark,
 		textDecoration: 'none',
--- a/src/components/GalleryPage/GalleryMenu.js
+++ b/src/components/GalleryPage/GalleryMenu.js
@@ -6,8 +6,25 @@ import Fade from '@material-ui/core/Fade';
 import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
 import { Link } from 'gatsby';
 import PropTypes from 'prop-types';
+import useTheme from '@material-ui/core/styles/useTheme';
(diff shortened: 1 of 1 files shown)


SubramaniamSatyen (approved):
Looks good!

src/components/GalleryPage/GalleryMenu.js
```diff
@@ -16,9 +16,17 @@ function GalleryMenu({ hothNames, galleryLinks }) {
 	const handleClose = () => {
 		setAnchorEl(null);
 	};
+	const linkStyles = {
```
khxia, line 19: You're right that there are better ways to add styling. For styling, we usually use material-ui's API to do styling. For example, take a look at `src/components/SchedulePage/Schedule.js` and how they use `makeStyles` and `useStyles` to do styling for an entire component. This also lets us access the `theme` object that we need for colors.

src/components/GalleryPage/GalleryMenu.js
```diff
@@ -16,9 +16,17 @@ function GalleryMenu({ hothNames, galleryLinks }) {
...
+		color: '#202020',
+		textDecoration: 'none'
+	};
+	const activeStyles = {
+		color: '#ff487e',
```
khxia, line 24: Instead of using hardcoded hexcolor values, we actually have a defined color palette through MUI in `src/components/layout.js`. Try to use those as much as possible instead. For example, this would be `theme.palette.secondary.main`

khxia (changes requested):

src/components/GalleryPage/GalleryMenu.js
```diff
@@ -6,8 +6,22 @@ import Fade from '@material-ui/core/Fade';
...
 import PropTypes from 'prop-types';
+import useTheme from '@material-ui/core/styles/useTheme';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles(theme => ({
+	linkStyle: {
+		color: theme.palette.primary.dark,
+		textDecoration: 'none',
+		'&:hover': {
+			color: theme.palette.primary.main
```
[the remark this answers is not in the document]
datowq: I think the background highlighting is enough. I removed the hover color change.

khxia (approved):
LGTM

PR #316, opened 2023-01-22
added "coming soon" message to workshop page

### LINKED ISSUE #302: Show workshops from last year
We will need to update the workshops page with our new workshops for HOTH X, but right now we're listing the workshops for HOTH 9. For now, add a message on the page like "Workshops coming soon, here are the ones from last year!" so that users can see an example of our HOTH workshops while knowing that they're from last year

Monicaaawa:
fixes #302

### DIFF
--- a/src/components/WorkshopPage/WorkshopPage.js
+++ b/src/components/WorkshopPage/WorkshopPage.js
@@ -34,9 +34,6 @@ const useStyles = makeStyles(theme => ({
 			paddingTop: theme.spacing(4)
 		},
 		paddingBottom: theme.spacing(4)
-	},
-	message: {
-		marginBottom: theme.spacing(1)
 	}
 }));
 
--- a/src/components/WorkshopPage/WorkshopPage.js
+++ b/src/components/WorkshopPage/WorkshopPage.js
@@ -232,7 +232,8 @@ function WorkshopPage() {
 				</Typography>
 
 				<Typography variant='body1' component='body'>
-					HOTH X workshops will be coming soon. Check out our previous workshops on our <a href="https://www.youtube.com/watch?v=GskV5qMOpkE&list=PLPO7_kXilXFbjHembKB9UdcMyFkc_1Q06">YouTube channel</a> if you are interested!
+					HOTH X workshops will be coming soon.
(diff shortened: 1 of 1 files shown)


jiangxuann (approved):

src/components/WorkshopPage/WorkshopPage.js
```diff
@@ -225,9 +226,15 @@ function WorkshopPage() {
...
 		<Container maxWidth='md' style={{ marginBottom: theme.spacing(8) }}>
-			<Typography variant='h4' component='h1' className={classes.title}>
-				Workshops
-			</Typography>
+			<Box style={{ marginBottom: theme.spacing(4) }}>
+				<Typography variant='h4' component='h1' className={classes.title}>
+					Workshops
+				</Typography>
+
+				<Typography variant='body1' component='body1'>
```
khxia, line 234: the `component` prop refers to the actually html tag or react component that is being used to render the text. In this case, I don't think that there is a body1 tag, so you should be using `body` instead.

khxia (changes requested):

src/components/WorkshopPage/WorkshopPage.js
```diff
@@ -225,9 +226,15 @@ function WorkshopPage() {
...
-			<Typography variant='h4' component='h1' className={classes.title}>
-				Workshops
-			</Typography>
+			<Box style={{ marginBottom: theme.spacing(4) }}>
+				<Typography variant='h4' component='h1' className={classes.title}>
+					Workshops
+				</Typography>
+
+				<Typography variant='body1' component='body'>
+					HOTH X workshops will be coming soon. Check out our previous workshops on our <a href="https://www.youtube.com/watch?v=GskV5qMOpkE&list=PLPO7_kXilXFbjHembKB9UdcMyFkc_1Q06">YouTube channel</a> if you are interested!
```
khxia, line 235: "Check out our previous workshops below from last year if you are interested!"

khxia (changes requested):

katelynsyu (approved):
lgtm

khxia (approved):
lgtm

PR #317, opened 2023-01-22
Removed schedule subtitle

### LINKED ISSUE #300: Make schedule unavailable
The schedule that is currently on the website will need to be updated, but let's first make it unavailable so that there is no schedule shown on the page. (I think it's a boolean on the Schedule component)

jiangxuann:
I removed the subtitle in schedule.

Fixes #300


katelynsyu (approved):
looking good :)

SubramaniamSatyen (approved):

PR #318, opened 2023-01-29
Updated FAQ section with date/time/location

jsi19:
Also created variables for date/time/location and concatenated into text.

### DIFF
--- a/src/components/HomePage/FAQSection.js
+++ b/src/components/HomePage/FAQSection.js
@@ -123,7 +123,7 @@ function FAQSection() {
 		},
 		{
 			question: `Who can join?`,
-			answer: `Applications for HOTH 9 are open to anyone! However, we are prioritizing 
+			answer: `Applications for HOTH X are open to anyone! However, we are prioritizing 
 				UCLA students for our prize tracks as well as in-person hacking.`
 		},
 		{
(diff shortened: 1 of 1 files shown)


SubramaniamSatyen (changes requested):
On line 152, should use the hothNum variable. Looks good otherwise

src/components/HomePage/FAQSection.js
```diff
@@ -110,6 +110,10 @@ AccordionQA.propTypes = {
 };
 
 function FAQSection() {
+	const hothNum = 'X';
+	const appDeadline = 'Friday, February 24, 2023';
+	const hothDate = 'Sunday, March 5, 2023';
```
katelynsyu, line 115: I really like this idea to have everything in a constant and just call the constant. Turns out we have a file already with constants in it, called constants.js under components. if you could move these constants to that file and update the constants there, that would be amazing

jsi19 (commented):
Fixed

katelynsyu (changes requested):

katelynsyu (changes requested):
On line 177, there's an extra exclamation mark that shouldn't be there

src/components/HomePage/FAQSection.js
jsi19: Will do when available if not assigned to other

jsi19 (commented):
Fixed

SubramaniamSatyen (approved):
If we're holding off on moving the variables to a shared place across the site, looks good!

katelynsyu (approved):
lgtm

PR #324, opened 2023-01-30
Change banner layout

katelynsyu:
We had to change the layout so that the text goes on top of the image rather than being a grid


jodymlin (approved):
lgmt 👍

maggieelli (approved):
lgtm

katelynsyu:
Screenshot 2023-01-30 at 9 35 04 AM
bruh I tried adding the countdown and I'd say we have an issue again lol. I will try to fix this later

maggieelli (changes requested):
Add new banner image in Assets folder :)

jodymlin (commented):
lgtm but failing some of the eslint checks

go ahead and run `yarn lint --fix` and it should fix those for you

src/components/HomePage/Banner.js
```diff
@@ -211,47 +185,80 @@ function Banner() {
...
-								<Box><Img fluid={data.placeholderImage.childImageSharp.fluid} /></Box>
-							</Box>
-						</Grid>
-					}
-				</Grid>
+				</Box>
+				{smallScreen ?
+					null :
+					<Box display='flex' flexDirection='column' className={classes.img}>
+						<Box><Img fluid={data.placeholderImage.childImageSharp.fluid} /></Box>
```
khxia, line 259: What is this <Box> for again? Was it so that the flexbox treats it as an item?

katelynsyu: The inner box around the `` is unnecessary, I think. I believe that was from when we had the count down in the outer box. I think we can delete it.

maggieelli: Deleted!

khxia (approved):
LGTM.
Banner code is never pretty haha. Maybe we can investigate the slightly blurry banner in another issue. Maybe we're stretching it. https://github.com/gatsbyjs/gatsby/issues/3984

maggieelli (approved):

PR #336, opened 2023-02-19
Added Sticky Nav Menu

SubramaniamSatyen:
The text can be customized. Spent a bit of time looking into getting the current section to be highlighted - I believe it's doable with the useRef() hook. Otherwise I believe changes add intended functionality


katelynsyu (changes requested):
can we play around with using primary and secondary colors for the buttons on the nav menu to offset all the dark text on the page? also is it possible to push the nav bar text down so when we're at the top of the page the nav bar text is aligned with the "workshop" title?

katelynsyu (changes requested):
Also it doesn't seem like you're using window.js, so maybe deleting the file is best? For future reference, you can use theme.breakpoints if you want to edit how a component looks based on how narrow a window is :)

SubramaniamSatyen:
I've gone ahead and added some extra styling and removed the window page. Feel free to let me know if anything still needs to be changed whenever you get the chance.

katelynsyu:
YAY!!! I think the functionality is perfect!!! Now we just have to work on editing how it looks. Is it possible to get rid of the indent and the bullet point for the sidebar? Also can we add a faint line on the right side of the sidebar to show a differentiation between the navbar and the rest of the screen? 

Honestly, I really don't know what would look better, so I'm just spit balling ideas. Play around with it and if you don't like my ideas, feel free to change them :). Thanks!!

SubramaniamSatyen:
Added a bit more styling, but let me know feel free to let me know if it still looks off or any additional thoughts

jakobreinwald (approved):
LGTM everything looks great, thanks for all the work back and forth with us. :)

katelynsyu (approved):
LGTM!

PR #337, opened 2023-02-19
Added datetime helper function

### LINKED ISSUE #213: Make a `datetime_utils` file
There seem to be many different places in our repo where we deal with DateTime. Sometimes, these places get their `DateTime` information in different formats, and they also need their `DateTime` string in different formats too. Some examples include:

- Workshop Schedule
- Announcements Page
- Home Announcements Banner
- Home Page Countdown

Some of these components already have their own DateTime helper functions. But perhaps it would be better if we had a central `datetime_utils.js` file where we can just use those helper functions that abstract away all the annoying stuff like formatting and timezones for us such that we don't have to worry about it when dealing with time. This way we would also have a consistent way of how we deal with DateTime instead of having different methods scattered throughout the repo.

Monicaaawa:
fixes #213
Wrote DateTime helper function in datetime_utils.js (under utils folder) and standardized the display format for date and time in announcement page, homepage banner, and homepage announcement banner.


src/components/HomePage/Banner.js
```diff
@@ -224,10 +224,10 @@ function Banner() {
 							style={{ marginBottom: 10, fontWeight: 500 }}
 							component='h3'
 						>
-							<time dateTime={hothStart.toISOString()} hidden>
-								{month} {startDay}{endDayString}, 2022
+							<b>Date: </b><time dateTime={hothStart.toISOString()} >
+								{month} {startDay}{endDayString}, 2023
 							</time>
-							<b>Date:</b> Sunday, March 5, 2023
+							{/* <b>Date:</b> Sunday, March 5, 2023 */}
```
khxia, line 230: If you are not using this code anymore, then you should delete it to keep our code clean.

src/utils/datetime_utils.js
```diff
@@ -0,0 +1,10 @@
+export function getDateTime(timestamp) {
+	try {
+		const d = new Date(timestamp);
+		const date = d.toLocaleDateString('en-US');
+		const time = d.toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit' });
+		return `${date}, ${time}`;
```
khxia, line 6: We shouldn't be returning a string with a specific format like `${date}, ${time}` hardcoded in.
Consider the following use cases:
- What if another component wants to use this format, but another component wants to display time like `${date} - ${time}` instead?
- Or what if a component doesn't care about the time, and only cares about the date?

This function should be able to handle either regardless. To fix this, we should instead return an object that maps keys to values so that each component can use the data however it wants. Like so:

```js
return {
  date: date,
   time: time
}
```

And conveniently, javascript lets a simplify this code into the following that does the exact same thing:
```js
return { date, time }
```

Then, in your components, you can use it like so:

```
const datetime = getDateTime(timestamp);
...

<p> Posted on {datetime.date}, {datetime.time} PST </p>
```

src/components/HomePage/Banner.js
```diff
@@ -224,10 +224,10 @@ function Banner() {
 							style={{ marginBottom: 10, fontWeight: 500 }}
 							component='h3'
 						>
-							<time dateTime={hothStart.toISOString()} hidden>
-								{month} {startDay}{endDayString}, 2022
+							<b>Date: </b><time dateTime={hothStart.toISOString()} >
+								{month} {startDay}{endDayString}, 2023
```
khxia, line 228: We should not be hardcoding the year, but also getting the year from the hothStart date.

src/utils/datetime_utils.js
```diff
@@ -0,0 +1,10 @@
+export function getDateTime(timestamp) {
```
khxia, line 1: Also, there seems to be a `timezone_names.js` file that has functionality similar to this file, so you should move the function in that file into this file and make this file the centralized util file that handles anything 'time' related.

khxia (changes requested):

ariyin (approved):
i trust alex 👍

src/utils/datetime_utils.js
```diff
@@ -0,0 +1,25 @@
+function getDateTime(timestamp) {
+	try {
+		const d = new Date(timestamp);
+		const date = d.toLocaleDateString('en-US');
+		const time = d.toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit' });
+		return { date, time };
+	} catch {
+		return '';
```
khxia, line 8: Return null instead, the default of an empty object

src/components/HomePage/Banner.js
```diff
@@ -24,6 +24,7 @@ import {
 
 // These dates are displayed in the user's timezone
 const monthFormatter = new Intl.DateTimeFormat('en-US', { month: 'short' });
+const dayOfWeek = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
```
khxia, line 27: Move this logic into datetime_utils by creating a function called "getDayOfWeek"

src/utils/datetime_utils.js
```diff
@@ -0,0 +1,25 @@
+function getDateTime(timestamp) {
```
khxia, line 1: In a util functions file, we should be documenting each function as much as possible so other people know how to use it. Add some comments here to describe the function. Make sure to mention what arguments it takes and what it returns.

khxia (changes requested):

PR #411, opened 2024-02-25
Kaylin/making nav bar buttons interesting

### LINKED ISSUE #375: Update nav bar, look at hack website for reference
We want to make the buttons more interesting and have cool animations on mouse hover

kaylin-chung:
Fixes #375


jakobreinwald (changes requested):
Hey Kaylin, everything looks great but I just have a few small changes and then we can get this merged in: 

- Can we increase the space between the Apply button and the rest of the page links
- When dragging the page size from right to left to shrink, you'll see that before condensing the page links into the hamburger button, the Apply button gets stacked under the other ones. We want to switch to the condensed/mobile version before this happens, so you'll have to increase the size of the breakpoint. 
- Can we switch the color of the Apply button back to what it currently is on the HOTH website, just because it stands out a bit more. 

Tyyy let me know if you have any questions

PR #442, opened 2024-03-10
hid schedule from AppBar

### LINKED ISSUE #439: Hide schedule

SPerrott22:
schedule page endpoint still up but users will not see it as it is no longer in navbar.

Fixes #439


SPerrott22:
Made scheduleAvailable depend on `constant.js`

katelynsyu:
@jakobreinwald do we want to hide the schedule from the nav bar? I thought we keep it in the nav bar and just have the `ComingSoon` component rendered on page.

PR #462, opened 2025-01-19
Fix redirect issues

danielhzhou:
Edited netlify.toml to include [[redirects]] to index.html to have react router to catch and redirect to the correct routes.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,3 +12,8 @@ NODE_VERSION = "20.17.0"
 
 [context.production]
 ignore = "true"
+
+[[redirects]]
+from = "/*"
+to = "/index.html"
+status = 200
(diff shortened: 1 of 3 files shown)


danielhzhou:
This fix does cause the path /(a page we dont have) to just be a blank white page, would you like me to make it an error page, page not found, etc?

ariyin:
> This fix does cause the path /(a page we dont have) to just be a blank white page, would you like me to make it an error page, page not found, etc?

yeah we can probably make a 404 error page by copying the one on the hack website and integrating it: https://github.com/uclaacm/hack.uclaacm.com/blob/main/src/pages/NotFound.jsx

danielhzhou:
> > This fix does cause the path /(a page we dont have) to just be a blank white page, would you like me to make it an error page, page not found, etc?
> 
> yeah we can probably make a 404 error page by copying the one on the hack website and integrating it: https://github.com/uclaacm/hack.uclaacm.com/blob/main/src/pages/NotFound.jsx

added!

ariyin (approved):
lgtm

jamesmwu (approved):
LGTM!

PR #463, opened 2025-01-19
Arnav/image lazy loading

aroy23:
Added lazy loading for all images and implemented custom lazy loading for the YouTube video player on the workshops page


aroy23:
I wasn't sure if you guys wanted the lazy loading for the video players under workshops. I pretty much made it load the thumbnail lazily and then the actual video gets loaded on the click of the thumbnail i can change it back if its unnecessary though.

ariyin:
it doesn't seem like anything has changed for me on both the deployed version and a local version of your branch. there's also a split second where when you refresh the home page banner doesn't take up any space at all and the text overlaps and you can kind of seeing the content get pushed down.

https://github.com/user-attachments/assets/dd402998-02a1-4569-9f4d-8fc4236c4373

also imo the thumbnail feature for the videos is worse ux since users might not know it's an actual video.

does it work properly for you?

aroy23:
The lazy loading does work on my end I've attached two videos showing the difference in how the images are loaded in my network tab for the original vs new implementations. I believe the overlapping banner issue may be an unrelated issue to lazy loading. 

I can also remove the thumbnail feature, but then the videos wouldn't be lazily loaded so there would be a maybe an extra quarter of a second extra wait time if that's okay.

Original: (Instantly Loaded)

https://github.com/user-attachments/assets/d5cb3e81-ff4a-4681-97c1-44a33a37e103

Lazy:

https://github.com/user-attachments/assets/5b880f0f-09ac-4aaa-9fc9-c8c5c9ab2b1e

ariyin:
i see what you mean. i think when james created the issue he also probably meant for there to be a transition when an image is lazily loaded so it doesn't come off lagging and like loading chunks of an image at a time. with lazy loading you can also take up space for the image before it's even fully loaded so we don't have the overlapping banner image. 

https://github.com/user-attachments/assets/09d08293-de09-4120-bc9f-3cab828fce73

i'll take this example from my own website where the frame still maintains its size even when the image isn't loaded and the image kind of fades into view while it loads. i'll leave this for james to look at since i'm only guessing his intentions and not rlly sure either.

src/components/Home/PhotoCarousel.jsx
```diff
@@ -12,6 +12,8 @@ import image5 from '../../images/carousel/image5.jpg';
 import image6 from '../../images/carousel/image6.jpg';
 import image7 from '../../images/carousel/image7.jpg';
 
+import { LazyLoadImage } from 'react-lazy-load-image-component';
```
jamesmwu, line 15: nit: Let's move this to be around line 6, still maintaining a space between it and the images.

I think it'd just for readability long term, we can keep the packages at the beginning and for this particular component since (currently) there's a bunch of images we have to manually import, those can be separated.

src/components/Workshops/Workshop.jsx
```diff
@@ -20,12 +22,29 @@ export default function Workshop({ title, youtube, author, description, readme,
...
+            )}
+            {isPlayerVisible && (
+                <ReactPlayer
+                    className='workshop-video'
+                    id={'id_' + title.replace(/ /g, '_')}
+                    controls={true}
+                    width='100%'
+                    url={youtube}
+                />
+            )}
```
jamesmwu, line 46: Let's update this-- as Jenna commented earlier, the lazy loading is great but the UX somewhat suffers. Looking at the deploy, I think it's hard to tell that it's a youtube video you can click on.

It'd be great if we could lazily load the current youtube embed that we have so it's still easily identifiable as a video.
image

jamesmwu (changes requested):
Hi Arnav, thanks for working on this! Looks great, site feeling smoother already :)

Just have some small review comments for you here. Additionally, like Jenna mentions earlier in this thread, it'd be great if we could have a smoother transition for the images while we lazily load them, as opposed to doing so in chunks. 

From a brief glance at the package you've chosen to go with here: https://www.npmjs.com/package/react-lazy-load-image-component
it seems like there are some built in things to help easily do this. I think it'd be great to do either blur or black and white, so it can be your call between the two of those which you think would be best to go with.

jamesmwu:
> I wasn't sure if you guys wanted the lazy loading for the video players under workshops. I pretty much made it load the thumbnail lazily and then the actual video gets loaded on the click of the thumbnail i can change it back if its unnecessary though.

For more clarity regarding this, I think if we could lazily load the thumbnail while still keeping the youtube controls, that'd be the most ideal. If you look into it for a while and there's no way to do it, then as a number two it'd still be better to keep it as it previously was than to make the UX unclear. Thanks for working on this!

jamesmwu (changes requested):
Awesome job Arnav!!! I think the only thing now is I'm noticing a bug on the workshops page, where the headings of each type of workshop are not aligned with the headers. Please take a look at this and I think after that, we're good to merge this in. Thanks!

image

aroy23:
Fixed! sorry about that.

ariyin:
a small nit:

image

there seems to be a grey line underneath the thumbnail, and it has to do with the background color styling. can we fix this so it doesn't show and the bottom of the thumbnail is still rounded?

jamesmwu (approved):
LGTM! Pending Jenna's fixes

hannahkendall04 (dismissed):
Hi Arnav! Great work!! The lazy loading looks SO good!

I have the same small nit picky thing as Jenna - I'm thinking it might look cleaner without the gray edges around the videos on the workshops page. From what I can see, should just involve changing lines 188 and 193 in the Workshops.css file to either be white or transparent. That being said, if you like the gray borders, it is more than OK to keep them in!

Regardless of what you would like to do with the borders, would you be able to fix the merge conflicts in the HothDescription.jsx file when you get the change? After that should be good to merge!

ariyin (approved):

PR #465, opened 2025-01-21
Apply Button

dcho-jaewook:
Added the apply button for the nav-bar

Included a boolean variable "IsApplyActive" and a string variable "ApplyButtonClassName" to make it easier to switch the state in the future


ariyin:
two things from me:
- i think the button would look better with white text (when it's not hovered)
- i don't think we should deploy this with the link linking to "/link". we can probably put the actual application link there and wait on deploying this for when we're ready to open applications (but i will let james confirm)

src/styles/Navbar.css
jamesmwu, line 1: Some additional nits:
- The spacing between the Apply button and the rest of the headers doesn't seem even. Please add some more spacing between the button and the rightmost `Gallery` button.
- Please change the bold the text inside the button, I think that might look better here(?). Additionally, let's make the background color the Hack accent color (the reddish pinkish color). All these styling changes are for the ACTIVE button-- as stated earlier, we should be deactivating this button since the form isn't live yet.

jamesmwu (changes requested):
Have some small changes here, and then should be good to go! Thanks Jaewook!

src/styles/Navbar.css
```diff
@@ -85,6 +85,18 @@
...
+	color: white;
+	background-color: var(--lightPurple);
+}
+
+a.apply-inactive {
+	pointer-events: none;
+	color: var(--gray);
+	background-color: var(--bgGray);
+}
+
```
jamesmwu, line 99: Please address the issues I raised previously. To recap:

- The spacing between the Apply button and the rest of the headers doesn't seem even. Please add some more spacing between the button and the rightmost Gallery button.
- When the button is active, let's make the background color the Hack accent color (the reddish pinkish color). `var(--hackAccent)`

jamesmwu (changes requested):

src/components/General/Navbar.jsx
```diff
@@ -7,6 +7,8 @@ import HackLogo from '../../images/hothXI-logo.svg';
 export default function Navbar() {
 	const [isOpen, setIsOpen] = useState(false);
 	const [isMobile, setIsMobile] = useState(false);
+	const IsApplyActive = false;
+	const ApplyButtonClassName = 'apply-btn ' + (IsApplyActive?'apply-active':'apply-inactive');
```
jamesmwu, line 11: Let's stick to camel case for these variables:
```
isApplyActive
applyButtonClassName
```

After that, lgtm!

jamesmwu (changes requested):

jamesmwu (approved):
LGTM!

ariyin:
do we want it to be active rn?

jamesmwu:
> do we want it to be active rn?

Nah, Lillian and Max told me to keep it inactive for now since they haven't yet opened for apps

ariyin:
> > do we want it to be active rn?
> 
> Nah, Lillian and Max told me to keep it inactive for now since they haven't yet opened for apps

in the current commit it's active

jamesmwu:
Ah shoot I missed that, my bad. @dcho-jaewook Please change the button to be inactive.

Before you publish the PR and rerequest review, please make sure that all the feedback previously given is still fixed so we don't need to repeat the same things over again. Thanks!

> > > do we want it to be active rn?
> > 
> > 
> > Nah, Lillian and Max told me to keep it inactive for now since they haven't yet opened for apps
> 
> in the current commit it's active

src/components/General/Navbar.jsx
```diff
@@ -7,8 +7,8 @@ import HackLogo from '../../images/hothXI-logo.svg';
 export default function Navbar() {
 	const [isOpen, setIsOpen] = useState(false);
 	const [isMobile, setIsMobile] = useState(false);
-	const IsApplyActive = false;
-	const ApplyButtonClassName = 'apply-btn ' + (IsApplyActive?'apply-active':'apply-inactive');
+	const isApplyActive = true;
```
jamesmwu, line 10: Please change this to false. Before rerequesting, make sure that all previous feedback is still fixed.

jamesmwu (changes requested):

dcho-jaewook:
For the last commit, I fixed the following errors
- set `isApplyActive` to false
- constant names on Navbar.jsx line 10 & 11 (changed to `applyButtonClassName` and `isApplyActive`)
- changed from `<Link>` tag to `<a>` tag on Navbar.jsx line 88
- removed bold style and changed color to `--hackAccent`on Navbar.css line 88 to 91

I'm not sure why but the 'Fixed previous errors' commit doesn't show the entire changes

ariyin:
can you resolve the merge conflicts?

jamesmwu (approved):
LGTM

PR #477, opened 2025-01-29
Nishant/section-fade-in

### LINKED ISSUE #471: Add section fade in
In an effort to modernize our website, let's have some sections fade in on scroll.

An example you can refer to is how this looks on the home page of the Hack website (although the hack website also still requires some tune-up for this particular feature). https://hack.uclaacm.com/

We should apply this to just our HOTH Home page for now, but later we could also discuss whether it'd be appropriate for our Announcements and Schedule page once those two are complete.

Nishant-Ray:
- Seems like the `react-animate-on-scroll` simply requires a `<link>` to a CSS file for it to work properly
- Home page sections should have the fade-in animations working properly so we can now add them to the other pages quite easily with the `<ScrollAnimation>` wrapper component
- Alternatively, another option would be to use the `aos` package to be consistent with the Hack website (https://github.com/uclaacm/hack.uclaacm.com/pull/496)

Fixes #471

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,6 +5,7 @@
 		<link rel="icon" type="image/svg+xml" href="/src/images/hothXI-logo.svg" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<title>Hack on the Hill</title>
+		<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/animate.css/3.5.2/animate.min.css">
 	</head>
 	<body>
 		<div id="root"></div>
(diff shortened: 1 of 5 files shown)


ariyin:
i ended up switching the hack website to the aos package bc i think it works a bit better, you can customize the scroll animations, and it doesn't require wrapping stuff in another component, so aos would probably be preferable

Nishant-Ray:
Got it! I'll switch to that then

Nishant-Ray:
I switched to AOS so feel free to review now @ariyin @jamesmwu

ariyin (approved):
lgtm, also forgot to mention originally you should add the original issue tag in your pr as well (#471)

Nishant-Ray:
Oh yeah thanks, sorry about that!

jamesmwu (approved):
LGTM! Thanks Nishant!

PR #481, opened 2025-01-31
Nishant/toggle theme

### LINKED ISSUE #470: Add light / dark mode
Let's add a button on the navbar, to the left of the "Home" page. It should be a picture of a sun / moon icon. When clicking it, it should toggle the theme of the website from light to dark.

It's important that for this particular feature, it should be applicable to ALL pages without breaking any existing styles, layouts, etc. We should specify the new background colors in the `:root` section.

Nishant-Ray:
### Overview
- Implemented a toggle button in the navbar to switch between light and dark mode across the entire site
- Fixes #470

### Changes
- Added package `react-switch` to get a customizable toggle switch
- Created `DarkModeToggle.jsx` accompanied with `DarkModeToggle.css`
    - Lets user dynamically change coloring theme by adding/removing CSS class `.dark` from `:root` document element
    - Uses `localStorage` to remember user's theme choice upon leaving the website
    - Created a custom toggle switch using new `react-switch` package
- Added the theme toggler to the navbar
    - To ensure that it fit properly with smaller screen sizes, had to change the "global" max mobile screen length from 950px to 1000px
- Added dark mode CSS variables in `src/styles/App.css` under `:root.dark`
    - Also added few more variables to have dynamic color changing based on selected theme
    - Modified CSS files of several pages to make theme changing compatible
    - To add light/dark mode coloring to the FAQ section chevron SVG, I had to add the SVG directly inline in `<AccordionItem>` and get rid of the `` implementation in `src/components/FAQSection.jsx`
        - This was the cleanest solution that I could come up with (I couldn't find figure out how to cleanly add dynamic recoloring to the SVG using the `` implementation after trying every possible method, like using a CSS mask, using different SVG files for light and dark mode, using CSS invert filter, and more)
        - If there's some other method that could work and is cleaner, let me know!

### New Bugs
- None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "react-player": "^2.16.0",
         "react-router-dom": "^7.0.2",
         "react-slick": "^0.30.3",
+        "react-switch": "^7.1.0",
         "slick-carousel": "^1.8.1"
       },
       "devDependencies": {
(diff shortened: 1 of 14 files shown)


hannahkendall04 (approved):
Super cool! Looks good to me!

jamesmwu:
This Looks great to me, but would like to hear @ariyin or @melissaqshi 's opinions on the Footer color-- looks somewhat out of place to me, but I don't know what replacement color would be better.

Suggestions would be great :0

ariyin:
i think `--hackSecondaryV3` might be a better footer color? also can we change the light/dark colors (blue/yellow) to match the hack theme (to pink/purple)? and maybe it'd be better to move the button itself to be the rightmost button on the navbar.

Nishant-Ray:
Yeah good ideas, I'll do that

Nishant-Ray:
How does it look now? @ariyin @jamesmwu

ariyin (approved):
i'm a fan

jamesmwu:
Looks good to me as well! Thanks Nishant!

PR #483, opened 2025-02-01
Created data/PrizeData to map prizes to the Prizes.jsx page

### LINKED ISSUE #475: Put prize data in src/data
Right now we've hardcoded the prize data into Prizes.jsx. Let's store the data instead in src/data using a new file named `PrizeData.js` or similar, and map it.

This is meant to make long term maintenance easier.

danielhzhou:
Fixes #475 

put prizeData in the format:

{
    id: 1,
    name: 'Best Overall',
    type: 'overall',
    label: 'PRIZE',
    amount: '$150 Amazon Gift Card'
},

Changed Prizes.jsx to display the exported prizeData array

Created helper function "getPrizeIcon" to fetch the correct prize icon based on the type

### DIFF
--- a/src/data/PrizeData.js
+++ b/src/data/PrizeData.js
@@ -1,0 +1,46 @@
+const prizeData = [
+{
+    id: 1,
+    name: 'Best Overall',
+    type: 'overall',
+    label: 'PRIZE',
+    amount: '$150 Amazon Gift Card'
+},
+{
+    id: 2,
+    name: 'Best Web',
+    type: 'web',
+    label: 'PRIZE',
+    amount: '$80 Amazon Gift Card'
+},
+{
+    id: 3,
+    name: 'Best Mobile',
+    type: 'mobile',
+    label: 'PRIZE',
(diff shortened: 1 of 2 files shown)


ariyin (changes requested):
is there a reason why we need an id, type, and label field in the json? can we not pull the image src from the name of the prize itself and have the map generate its own key? so like:

```
{prizeData.map((prize, id) => (
     <div className='prize-card' key={id}>
```

and also i think the original purpose of the label "prize" was just to indicate it's a prize, so i'm not sure if we need a label field. 

^^ these are just suggestions, the only actual fix is that the amount for best mobile should be $80. 

also reminder to make sure to tag the original issue like #475 in the pr itself

danielhzhou:
@ariyin yes you are right, i simplified the data down to name and amount, used the name as the key and mapped the correct image from the name. 

also fixed the background grey circle + got rid of the label

ariyin (approved):

jamesmwu (approved):
LGTM!

PR #501, opened 2025-02-09
melissa/fixed-workshop-page-buttons

### LINKED ISSUE #478: fix buttons on workshop page
either:
- align the buttons on the bottom for each row
- move the buttons to go below the header of the title and instructor (i think i'd prefer this)

also:
- get rid of the text and find a better slides icon
- here's my suggestion for placement: left align the icons and give it a set gap instead of using justify-between 

Image

melissaqshi:
## Overview
- Fixed the buttons on the Workshop Page
- Fixes #478 

## All Changes
- Deleted the text and found a different slides icon 
- Moved the buttons from below the description to above the description


src/components/Workshops/Workshop.jsx
```diff
@@ -1,7 +1,7 @@
 import React, {useState} from 'react';
 import '../../styles/Workshops.css';
 import ReactPlayer from 'react-player/youtube';
-import { Github, Youtube, File } from '@geist-ui/icons';
```
Nishant-Ray, line 4: What about the `Monitor` or `Display` icons for the slides icon? I feel like those look a little better

src/styles/Workshops.css
```diff
@@ -103,12 +107,13 @@
     letter-spacing: .5px;
     font-weight: 300;
     margin-top: 0;
-    margin-bottom: 0;
+    margin-bottom: 0.2em;
```
Nishant-Ray, line 110: Try to use `rem` instead of `em` whenever you can

src/styles/Workshops.css
```diff
@@ -141,6 +147,7 @@
     }
 
     .workshop-description {
+        margin-block-start: 0px;
```
Nishant-Ray, line 150: Super tiny thing that doesn't really matter but again could you use `rem` instead of `px`

src/styles/Workshops.css
```diff
@@ -124,6 +129,7 @@
     padding: .5rem 1rem;
     margin-left: 1rem;
     margin-right: 1rem;
+    flex-direction: column;
 }
 
 .workshop-button:hover {
```
ariyin, line 135: can you add `cursor: pointer` to this? (i think this is the right thing but i'm not 100% sure, it should just make it so when you hover the workshop-button the cursor turns into a pointer)

src/styles/Workshops.css
```diff
@@ -103,12 +107,13 @@
     letter-spacing: .5px;
     font-weight: 300;
     margin-top: 0;
-    margin-bottom: 0;
+    margin-bottom: 0.2em;
 }
 
 .workshop-links {
     display: flex;
     justify-content: center;
```
ariyin, line 115: remove this

src/styles/Workshops.css
```diff
@@ -124,6 +129,7 @@
     padding: .5rem 1rem;
     margin-left: 1rem;
     margin-right: 1rem;
+    flex-direction: column;
 }
```
ariyin, line 133: in `.workshop-button` i think the only styling you need is:

```
background-color: var(--bgWhite);
border: none;
```

ariyin (changes requested):
after you fix the styling the buttons should probably look somewhat like this:
image

notice that there's still a green border around the icon. it's probably a margin/padding from some css class. i can't really figure out which class it belongs to thru inspect element, but if you can get rid of that, that'd be great too. and then increase the gap between the icons in `.workshop-links` to account for the deleted margin/padding.

Nishant-Ray (approved):
LGTM

ariyin (approved):
great job w all the changes!

PR #502, opened 2025-02-12
fix projects page #491

### LINKED ISSUE #491: fix projects page
- make thumbnails the same size
- align the button to the left
- make sure when you hover the **button** the cursor becomes a pointer, not only when you hover the **text**
- whatever other issues there are
- also technically the page should be updated (#437)

kiankian:
## Overview
- Made thumbnails the same size, and fixed hover and alignment of the button
- Fixes #491 

## All Changes
1) Made the thumbnails the same size by changing the max-height: 300px to height: 300px. Also, centered the object-position so the image is in the center
2) Changed the margin-left of the "see project" button so it is aligned to the left of each winner-info div
3) Changed the cursor to pointer anytime it hovers over the "see project" button rather than the text inside the button

## New Bugs/Issues
- N/A

## Other information
- N/A

### DIFF
--- a/src/styles/Gallery.css
+++ b/src/styles/Gallery.css
@@ -12,13 +12,14 @@
 .winner-image {
     border-radius: var(--rounded-lg);
     overflow: hidden;
-    max-height: 300px;
+    height: 300px;
 }
 
 .winner-image img {
     width: 100%;
     height: 100%;
     object-fit: cover;
+    object-position: center;
 }
 
 .winner-info {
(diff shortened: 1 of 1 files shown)


src/styles/Gallery.css
```diff
@@ -55,7 +56,7 @@
     border: none;
     border-radius: var(--rounded-md);
     padding: .5rem 1rem;
-    margin-left: 1rem;
+    margin-left: 0rem;
     margin-right: 1rem;
```
jamesmwu, line 60: Can we combine these two just using `margin: x x x x` syntax?

jamesmwu (dismissed):
LGTM for the most part but I would like @ariyin's confirmation if the gallery page now looks like what we expect

hannahkendall04 (approved):
LGTM! Won't merge so @ariyin can make comments if she would like.

ariyin (approved):

PR #532, opened 2025-03-02
FIxed issue past winner button should hover

kiankian:
Made the mouse into a cursor when you hover over the past winners button


jamesmwu (approved):
LGTM!

